PR #1, opened 2018-05-11
push to the empty branch

cizezsy:


pom.xml
```diff
@@ -0,0 +1,118 @@
...
+            <groupId>net.sf.trove4j</groupId>
+            <artifactId>trove4j</artifactId>
+            <version>3.0.3</version>
+            <type>jar</type>
+        </dependency>
+    </dependencies>
+
+    <!-- If you want this to appear on the wiki page:
+    <developers>
+      <developer>
```
jeffpearce: You should fill this out and put in your contact info

src/main/java/io/jenkins/plugins/coverage/CoverageAction.java
```diff
@@ -0,0 +1,81 @@
...
+    @CheckForNull
+    @Override
+    public String getIconFileName() {
+        return null;
+    }
+
+    @CheckForNull
+    @Override
+    public String getDisplayName() {
+        return "coverage report";
```
jeffpearce: I might be wrong, but I think this string shows up in the UI - if so, it should be made localizable

pom.xml
```diff
@@ -0,0 +1,118 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>plugin</artifactId>
+        <version>3.4</version>
```
oleg-nenashev: Better to use the latest version, 3.9 iirc

src/main/java/io/jenkins/plugins/coverage/BuildUtils.java
```diff
@@ -0,0 +1,17 @@
+package io.jenkins.plugins.coverage;
+
+import hudson.model.Run;
+
+public class BuildUtils {
+	public static Run<?, ?> getPreviousNotFailedCompletedBuild(Run<?, ?> b) {
```
oleg-nenashev: In Jenkins project we usually spaces instead of tabs for spacing. It's totally up to you, but tabs may be preferable for other contributors

src/main/java/io/jenkins/plugins/coverage/CoverageAction.java
```diff
@@ -0,0 +1,81 @@
...
+import org.kohsuke.stapler.StaplerProxy;
+
+import javax.annotation.CheckForNull;
+import java.lang.ref.WeakReference;
+import java.util.Collection;
+
+public class CoverageAction implements StaplerProxy, SimpleBuildStep.LastBuildAction, RunAction2 {
+
+    private Run<?, ?> owner;
+    private WeakReference<CoverageResult> report;
```
oleg-nenashev: Such configuration won't be serializable in 2.102+ IIRC. Need to check it

src/main/java/io/jenkins/plugins/coverage/CoverageAction.java
```diff
@@ -0,0 +1,81 @@
...
+import jenkins.tasks.SimpleBuildStep;
+import org.kohsuke.stapler.StaplerProxy;
+
+import javax.annotation.CheckForNull;
+import java.lang.ref.WeakReference;
+import java.util.Collection;
+
+public class CoverageAction implements StaplerProxy, SimpleBuildStep.LastBuildAction, RunAction2 {
+
+    private Run<?, ?> owner;
```
oleg-nenashev: make the field transient. It does not need to be serialized

src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
```diff
@@ -0,0 +1,87 @@
...
+            super(CoveragePublisher.class);
+            load();
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public boolean configure(StaplerRequest req, JSONObject json) throws FormException {
+            save();
```
oleg-nenashev: why do invoke `save()` before super method? With such approach data on the disk may differ from the memory model

src/main/java/io/jenkins/plugins/coverage/adapter/XMLCoverageReportAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+import java.io.File;
+import java.net.URISyntaxException;
+
+public abstract class XMLCoverageReportAdapter extends CoverageReportAdapter {
+
+    public XMLCoverageReportAdapter(String path) {
+        super(path);
+    }
+
+    public abstract String getXSL();
```
oleg-nenashev: Would be great to document interfaces. E.g. can XSD return `null`? If yes, add `@CheckForNull` annotation and clarify the behavior in Javadoc

src/main/java/io/jenkins/plugins/coverage/adapter/XMLCoverageReportAdapter.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public Document convert(File source) {
+        File xsl = getRealXSL();
+        return XMLUtils.getInstance().convertToDocumentWithXSL(xsl, source);
+    }
+
+    @SuppressWarnings("WeakerAccess")
+    protected File getRealXSL() {
+        try {
+            return new File(getClass().getResource(getXSL()).toURI());
```
oleg-nenashev: `getResource()` may return null, but it's not checked here

src/main/java/io/jenkins/plugins/coverage/exception/ConversionException.java
```diff
@@ -0,0 +1,23 @@
+package io.jenkins.plugins.coverage.exception;
+
+public class ConversionException extends RuntimeException {
```
oleg-nenashev: Please avoid using `RuntimeException`s in the production code. It would be better to make it a common exception and define interfaces and error propagation correctly

src/main/java/io/jenkins/plugins/coverage/adapter/CoverageReportAdapter.java
```diff
@@ -0,0 +1,43 @@
...
+import hudson.model.Describable;
+import hudson.model.Descriptor;
+import io.jenkins.plugins.coverage.targets.CoverageResult;
+import jenkins.model.Jenkins;
+import org.w3c.dom.Document;
+
+import java.io.File;
+
+
+public abstract class CoverageReportAdapter implements ExtensionPoint, Describable<CoverageReportAdapter> {
```
jeffpearce: We should group declarations together - I think Jenkins projects in general use sun coding conventions http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-141855.html#1852 - but @oleg-nenashev can verify

src/main/java/io/jenkins/plugins/coverage/BuildUtils.java
jeffpearce: I think we should have javadoc comments throughout

jeffpearce:
Chartable.java, CoverageChart.java, and Ratio.java seem like they belong in the `io.jenkins.plugins.coverage.targets` namespace - what do others think?

cizezsy:
Thanks for @jeffpearce and @oleg-nenashev 's comment! I will fix them.

src/main/java/io/jenkins/plugins/coverage/adapter/CoverageReportAdapter.java
cizezsy: What declarations should be grouped? I don't understand...Could you show me an example?

jeffpearce: I was mostly referring to variable and constructor order - here's what that doc says:

- instance variables - First public, then protected, then package level (no access modifier), and then private.
- Constructors
- These methods should be grouped by functionality rather than by scope or accessibility. For example, a private class method can be in between two public instance methods. The goal is to make reading and understanding the code easier.

oleg-nenashev:
+1 for merging for now. We can continue code reviews in merged PRs as well

PR #3, opened 2018-05-16
Add global thresh hold configuration, fix some bugs, improve Java docs.

cizezsy:


src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
```diff
@@ -13,22 +13,27 @@
...
 import org.kohsuke.stapler.StaplerRequest;
 
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import java.io.IOException;
 
 public class CoveragePublisher extends Recorder implements SimpleBuildStep {
 
     private CoverageReportAdapter[] adapters;
+    private ThreshHold[] globalThreshHolds;
```
oleg-nenashev: BTW, it is better to use lists instead of the arrays for the new code

src/main/java/io/jenkins/plugins/coverage/adapter/XMLCoverageReportAdapter.java
```diff
@@ -36,22 +36,27 @@ public XMLCoverageReportAdapter(String path) {
...
+            e.printStackTrace();
+            throw new ConversionException(e);
+        }
     }
 
     @SuppressWarnings("WeakerAccess")
-    protected File getRealXSL() throws FileNotFoundException {
+    protected File getRealXSL() throws ConversionException, FileNotFoundException {
         try {
             File realXSL = new File(getClass().getResource(getXSL()).toURI());
```
oleg-nenashev: `getXSL()` is marked as Nullable, but you do not check it here. I would suggest annotationg `getXSD` as `@CheckForNull` instead of Nullable. It would enforce FindBugs to check it and fail the build if it happens.

In the case of this method likely it needs to return null in this case as well

src/main/java/io/jenkins/plugins/coverage/adapter/XMLCoverageReportAdapter.java
```diff
@@ -36,22 +36,27 @@ public XMLCoverageReportAdapter(String path) {
...
-            if(!realXSL.exists()) {
+            if (!realXSL.exists()) {
```
oleg-nenashev: It would also make sense to check whether the `File` is actually a file (e.g. it may be pointing to a directory). there is `IsFile()` method

src/main/java/io/jenkins/plugins/coverage/adapter/util/XMLUtils.java
```diff
@@ -51,13 +49,13 @@ public Document convertToDocumentWithXSL(File xsl, File source) {
      * @param source Source file
      * @param result Result that want to be written in
      */
-    private void convertWithXSL(File xsl, File source, Result result) {
+    private void convertWithXSL(File xsl, File source, Result result) throws FileNotFoundException, ConversionException {
         if (!xsl.exists()) {
-            throw new ConversionException("XSL File not exist!");
+            throw new FileNotFoundException("XSL File not exist!");
```
oleg-nenashev: "does not"

src/main/resources/io/jenkins/plugins/coverage/CoverageAction/summary.jelly
```diff
@@ -2,7 +2,7 @@
 <j:jelly xmlns:j="jelly:core" xmlns:d="jelly:define" xmlns:l="/lib/layout" xmlns:t="/lib/hudson">
     <t:summary icon="graph.png">
         <div><a href="coverage/">${%Coverage Reports}</a></div>
-        <j:set var="result" value="${it.target}" />
+        <j:set var="result" value="${it.threshTarget}" />
```
oleg-nenashev: This object exists for `ThreshHold`, but not for `CoverageAction`. Not sure how it works

src/main/resources/io/jenkins/plugins/coverage/CoveragePublisher/config.jelly
```diff
@@ -4,4 +4,61 @@
         <f:hetero-list name="adapters" descriptors="${descriptor.listCoverageReportAdapterDescriptors}"
                        items="${instance.adapters}" hasHeader="true"/>
     </f:block>
+
+    <f:advanced>
```
oleg-nenashev: copied code as well, right?

cizezsy: Yeah, I copy it from Cobertura plugin. Should I add the license in Jelly file?

src/main/java/io/jenkins/plugins/coverage/adapter/XMLCoverageReportAdapter.java
cizezsy: I have marked `getXSL()` with `@CheckForNull`. Do you mean that I should check the return value of `getXSL()` in `getRealXSL()`?

cizezsy:
@oleg-nenashev Thanks, Oleg. I will run CheckStyle and FindBugs in my IDE to check them all.

src/main/java/io/jenkins/plugins/coverage/adapter/XMLCoverageReportAdapter.java
oleg-nenashev: yes.

cizezsy: I understand. Thanks~

abishek3876:
Is this a new plugin which will support getting coverage reports from multiple coverage tools? Which tools are you planning to support?
I wrote something similar [here](https://github.com/abishek3876/coverage-publisher), but that's in a very bad shape. Feel free to take pieces from there if you want.

cizezsy:
@abishek3876 Thanks! It will support multiple coverage tools. For now, it supports Jacoco and Cobertura, and I haven't decided which other tools I should implement yet. Once I decided, I will list them in README.

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -1,77 +1,172 @@
...
 
     private static final String DEFAULT_REPORT_SAVE_NAME = "coverage-report.xml";
 
-    public CoverageResult getCoverageReport(@Nonnull Run<?, ?> run, @Nonnull FilePath workspace, @Nonnull TaskListener listener, CoverageReportAdapter[] adapters) throws IOException, InterruptedException {
-        List<CoverageResult> results = convertToResults(run, workspace, listener, adapters);
+    private Run<?, ?> run;
+    private FilePath workspace;
+    private TaskListener listener;
+    private List<CoverageReportAdapter> adapters;
+    private List<ThreshHold> globalThreshHolds;
```
oleg-nenashev: Better to keep "H" in lower case. This is a single word

src/main/java/io/jenkins/plugins/coverage/threshhold/ThreshHold.java
```diff
@@ -0,0 +1,53 @@
+package io.jenkins.plugins.coverage.threshhold;
+
+import hudson.ExtensionPoint;
+import hudson.model.Describable;
+import hudson.model.Descriptor;
+import io.jenkins.plugins.coverage.targets.CoverageMetric;
+import jenkins.model.Jenkins;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+public class ThreshHold implements ExtensionPoint, Describable<ThreshHold> {
```
oleg-nenashev: Better to use "Threshold" here and in all occurrences

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -1,77 +1,172 @@
...
+
+
+    private boolean enableAutoDetect;
```
oleg-nenashev: As we discussed at the meeting, it would be great to move it into an extension point implementation if possible

src/main/java/io/jenkins/plugins/coverage/targets/CoverageElement.java
```diff
@@ -29,7 +29,7 @@
  * @since 22-Aug-2007 18:36:01
  */
 public enum CoverageElement {
-    AGGREGATED_REPORT(()->"Aggregated Report"),
+    AGGREGATED_REPORT(() -> "Aggregated Report"),
```
oleg-nenashev: As we discussed at the call, maybe it's better to convert it to "private final String name" instead of using Lambdas. But maybe there are some extensibility plans for the future.  CC @jeffpearce

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
cizezsy: I will do this in another pull request.

oleg-nenashev:
@cizezsy Make sure to merge the master PR into this branch so that you get the PR builder running

PR #4, opened 2018-05-22
[JENKINS-51368] Modernize report

cizezsy:
- [x] improve summary chart.
- [x] improve child sumary chart.
- [x] improve trend chart.
- [ ] improve source code chart.
- [ ] improve javascript code.
- [ ] improve jelly file.

Current Progress:

screenshot from 2018-05-22 20-37-36

screenshot from 2018-05-22 20-48-47


src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
```diff
@@ -183,7 +191,7 @@ private File getSourceFile() {
      * Getter for property 'sourceFileAvailable'.
      *
      * @return Value for property 'sourceFileAvailable'.
-     */
+     `*/
```
christ66: Unnecessary mark.

src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
```diff
@@ -435,22 +448,20 @@ public void setOwner(AbstractBuild<?, ?> owner) {
...
+            Run<?, ?> prevBuild = BuildUtils.getPreviousNotFailedCompletedBuild(owner);
+            CoverageAction action = null;
+            while ((prevBuild != null) && (null == (action = prevBuild.getAction(CoverageAction.class)))) {
+                prevBuild = BuildUtils.getPreviousNotFailedCompletedBuild(prevBuild);
+            }
+            return action == null ? null : action.getResult();
+        } else {
+            CoverageResult prevParent = parent.getPreviousResult();
+            return prevParent == null ? null : prevParent.getChild(name);
+        }
```
christ66: The commented out method, and this new added method look exactly the same (except for `CoverageAction` replacing `CoberturaBuildAction`).

cizezsy: I don't know exactly what you mean. I think the commented out code has already been replaced by new added code in this pull request. Did I understand it wrong?

christ66: Minor NIT, but if you want you can remove that commented part above. There's no need to have it.

cizezsy: I think the commented part above is already been removed in there....

oleg-nenashev:
In addition to the comment above, maybe we want to recover the color-blind story a bit. The current layout seems to be not that colorblind-friendly, maybe we could address it during one of the next phases

WDYT @cizezsy @jeffpearce ?

cizezsy:
@oleg-nenashev I agree.

jeffpearce:
@oleg-nenashev I agree, but I wonder if it's possible to do it on a per-user basis? I think that would allow it support scenarios like different users with different types of color blind issues.

oleg-nenashev:
@jeffpearce @cizezsy Let's discuss it at one of the next meeting. I agree it should be a User-based setting, and maybe it should be actually a part of the Jenkins core (Generic flags and API). I will create a ticket

PR #5, opened 2018-05-25
[JENKINS-51363] Pipeline support

cizezsy:
- add @Symbol on CoveragePublisher
- remove optional params in @DataBoundConstructor constructor, and using @DataBoundSetter public propery setter to config them.
- fix bug in XMLCoverageReportAdapter, which when plugin is packaged, xsl will not be properly find.


src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
```diff
@@ -64,23 +65,36 @@ public BuildStepMonitor getRequiredMonitorService() {
...
         return autoDetectPath;
     }
 
     @DataBoundSetter
-    public void setAutoDetectPath(String autoDetectPath) {
+    public void setAutoDetectPath(@Nonnull String autoDetectPath) {
         this.autoDetectPath = autoDetectPath;
     }
 
+    @Symbol("coverage")
```
oleg-nenashev: 'publishCoverage'?

cizezsy:
@oleg-nenashev Thresh branch is based on this branch, so I think I should not do too many changes in this branch. I will add more unit test in Thresh branch.

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -107,6 +107,7 @@ public CoverageResult processCoverageReport() throws IOException, InterruptedExc
 
         // If enable automatically detecting, it will try to find report and correspond adapter.
         if (isEnableAutoDetect()) {
+            listener.getLogger().println("Auto Detect is enable: Start to detect report");
```
oleg-nenashev: 1) enabled.
2) Looking for reports (maybe)

src/test/java/io/jenkins/plugins/coverage/CoveragePublisherPipelineTest.java
```diff
@@ -0,0 +1,169 @@
...
+        WorkflowRun r = project.scheduleBuild2(0).waitForStart();
+
+        Assert.assertNotNull(r);
+
+        j.assertBuildStatusSuccess(j.waitForCompletion(r));
+        j.assertLogContains("A total of 2 reports were found", r);
+
+    }
+
+    public static class CoverageScriptedPipelineScriptBuilder {
```
oleg-nenashev: Maybe it was an overkill. It could be preferable to just have scripts, which can be later used for documentation purposes

src/test/java/io/jenkins/plugins/coverage/CoveragePublisherPipelineTest.java
```diff
@@ -0,0 +1,169 @@
...
+                }
+            }
+
+            //TODO condition when adapter do not have @Symbol annotation.
+            return "";
+        }
+
+        private String generateSnippetForThreshold(Threshold threshold) {
+
+            return "[healthyThresh:" + threshold.getHealthyThresh()
```
oleg-nenashev: "thresh" is probably a bad wording for UI and API. It has different meaning (@jeffpearce please correct me if I am wrong). Does it come from Cobertura plugin? If no, I would recommend using a full "threashold" everywhere in API and Symbols

src/main/java/io/jenkins/plugins/coverage/adapter/CoberturaReportAdapter.java
```diff
@@ -35,6 +36,7 @@ public String getXSD() {
         return null;
     }
 
+    @Symbol("coberturaAdapter")
```
oleg-nenashev: maybe just `cobertura()` ? "Adapter" is an implementation detail

src/main/java/io/jenkins/plugins/coverage/adapter/JacocoReportAdapter.java
```diff
@@ -26,6 +27,7 @@ public String getXSD() {
         return null;
     }
 
+    @Symbol("jacocoAdapter")
```
oleg-nenashev: `jacoco`?

src/test/java/io/jenkins/plugins/coverage/CoveragePublisherPipelineTest.java
cizezsy: Yeah, I will change them in threshold branch.

cizezsy:
I have merged this pull request, and I will do all further changes in thresh branch.

PR #6, opened 2018-05-25
[JENKINS-51364] Adopt Threshold conditions (e.g. from Cobertura) so t…

cizezsy:
…hat the reporter can fail the build depending on conditions.

- add global threshold for all coverage report.
- add local threshold for each coverage report
- add **fail if unhealthy** option
- add health report

(This branch is based on pipeline branch... I have change thresh related code in pipeline branch, so for convenience, I create a new branch base on that.)


src/main/resources/io/jenkins/plugins/coverage/adapter/CoverageReportAdapterDescriptor/config.jelly
```diff
@@ -10,4 +10,49 @@
             </tr>
         </table>
     </f:entry>
+
+    <f:advanced>
```
oleg-nenashev: Add reference to the source implementation. License header may also need to be adjusted a bit

src/main/resources/io/jenkins/plugins/coverage/adapter/CoverageReportAdapterDescriptor/config.jelly
```diff
@@ -10,4 +10,49 @@
...
+        <f:entry title="Threshold">
+            <f:repeatable var="threshHold" name="thresholds" items="${instance.thresholds}">
+                <table width="100%">
```
oleg-nenashev: Since `Threshold` is describable, I would suggest moving its config to a `config.jelly` resource there an using `f:repeatableProperty` instead. Example: https://github.com/jenkinsci/github-branch-source-plugin/blob/900ea989a624223c88bddb3f1fbb3a3bd2193b5f/src/main/resources/org/jenkinsci/plugins/github_branch_source/GitHubConfiguration/config.jelly

src/main/java/io/jenkins/plugins/coverage/threshold/Threshold.java
```diff
@@ -1,49 +1,85 @@
...
 
 public class Threshold implements ExtensionPoint, Describable<Threshold> {
+
     private final CoverageMetric threshTarget;
 
-    private final float healthyThresh;
-    private final float unstableThresh;
-    private final float unhealthyThresh;
+    // mark build as unstable when coverage is less than this
+    private float unstableThresh = 0.0f;
```
oleg-nenashev: as suggested in parallel PR, better to use "threshold" instead of "thresh" everywhere

src/main/java/io/jenkins/plugins/coverage/CoverageAction.java
```diff
@@ -55,7 +60,7 @@ private CoverageResult getResult() {
 
         CoverageResult r = null;
         try {
-            r = CoverageProcessor.recoverReport(owner);
+            r = CoverageProcessor.recoverCoverageResult(owner);
         } catch (IOException | ClassNotFoundException e) {
```
jeffpearce: nit: I'm not a fan of one letter variable names -I'd be tempted to call this `coverageResult` instead of `r` - I find that easier to read for people new to the code base.

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -33,40 +40,44 @@
...
         this.listener = listener;
-        this.adapters = adapters;
-        this.globalThresholds = globalThresholds;
     }
 
 
-    public CoverageResult processCoverageReport() throws IOException, InterruptedException {
-        List<CoverageResult> results = convertToResults(adapters);
-        CoverageResult report = aggregatedToReport(results);
+    public void performCoverageReport(List<CoverageReportAdapter> adapters, List<Threshold> globalThresholds) throws IOException, InterruptedException, CoverageException {
```
jeffpearce: I'd really like to see javadoc comments for at least all the public methods

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -107,6 +118,7 @@ public CoverageResult processCoverageReport() throws IOException, InterruptedExc
 
         // If enable automatically detecting, it will try to find report and correspond adapter.
         if (isEnableAutoDetect()) {
+            listener.getLogger().println("Auto Detect is enabled: Looking for reports...");
```
jeffpearce: These end up going to the build log - I wonder if we should use java logging instead?

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -232,13 +348,36 @@ private CoverageResult aggregatedToReport(List<CoverageResult> results) {
...
+    public void setAutoDetectPath(String autoDetectPath) {
         this.autoDetectPath = autoDetectPath;
     }
 
     public boolean isEnableAutoDetect() {
-        return enableAutoDetect;
+        return !StringUtils.isEmpty(autoDetectPath);
+    }
+
+    public boolean isFailUnhealthy() {
```
jeffpearce: This is a property getter? It should probably be named `getFailUnhealthy`

jeffpearce: Same comment for the getters below

src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
```diff
@@ -64,20 +74,57 @@ public BuildStepMonitor getRequiredMonitorService() {
...
         return autoDetectPath;
     }
 
     @DataBoundSetter
-    public void setAutoDetectPath(String autoDetectPath) {
+    public void setAutoDetectPath(@Nonnull String autoDetectPath) {
         this.autoDetectPath = autoDetectPath;
     }
 
+    public boolean isFailUnhealthy() {
```
jeffpearce: get...

src/main/java/io/jenkins/plugins/coverage/adapter/CoberturaReportAdapter.java
```diff
@@ -35,6 +36,7 @@ public String getXSD() {
         return null;
     }
 
+    @Symbol("cobertura")
```
jeffpearce: I wonder if we need a new symbol here, so it doesn't conflict with the cobertura plugin? @oleg-nenashev - what do you think?

src/main/java/io/jenkins/plugins/coverage/adapter/JacocoReportAdapter.java
```diff
@@ -26,6 +27,7 @@ public String getXSD() {
         return null;
     }
 
+    @Symbol("jacoco")
```
jeffpearce: Potentially need a new symbol name here too

src/main/java/io/jenkins/plugins/coverage/adapter/CoberturaReportAdapter.java
jeffpearce: Ah - I see this change was suggested by @oleg-nenashev in a previous review, please disregard.

src/main/java/io/jenkins/plugins/coverage/adapter/JacocoReportAdapter.java
jeffpearce: Can also disregard

src/main/java/io/jenkins/plugins/coverage/adapter/CoberturaReportAdapter.java
cizezsy: Maybe we should add a simple prefix or suffix for this.

jeffpearce: I was mostly concerned it might somehow interfere with the existing cobertura plugin in the snippet builder, but if that's not a problem, I think it's OK.

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
christ66: Might be something for discussion, but even having a logger for this might not be useful since the java logging most likely gets flooded with other logs (would be difficult to find if debugging something). Maybe having a "debug" option 

For now I'd recommend putting `listener.getLogger()` into a variable and removing duplicate method calls everywhere else which would help with future improvements.

jeffpearce: You can set up different types of loggers in Jenkins though, so it's possible to configure it to see only jogs from this plugin.

jeffpearce: e.g. https://wiki.jenkins.io/display/JENKINS/Logging

src/main/java/io/jenkins/plugins/coverage/adapter/CoberturaReportAdapter.java
oleg-nenashev: There should be no conflicts between symbols as long as they declared for different extension points

src/main/java/io/jenkins/plugins/coverage/CoverageAction.java
```diff
@@ -1,25 +1,27 @@
 package io.jenkins.plugins.coverage;
 
-import hudson.model.Action;
-import hudson.model.HealthReport;
-import hudson.model.HealthReportingAction;
-import hudson.model.Run;
+import hudson.model.*;
```
oleg-nenashev: Better to avoid that in the production code. It may lead to import ambiguity in some edge cases

src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
```diff
@@ -13,46 +14,55 @@
...
+        processor.setFailUnhealthy(failUnhealthy);
+        processor.setFailUnstable(failUnstable);
+        processor.setFailNoReports(failNoReports);
 
-        CoverageAction action = new CoverageAction(result);
-        run.addAction(action);
+        try {
+            processor.performCoverageReport(getAdapters(), globalThresholds);
+        } catch (CoverageException e) {
+            listener.getLogger().println(e.getMessage());
```
christ66: nit: It would be nice to log the full stack trace in a logger. The message for the user is good.

src/main/java/io/jenkins/plugins/coverage/threshold/Threshold.java
```diff
@@ -1,41 +1,76 @@
...
     }
 
-    public float getUnstableThresh() {
-        return unstableThresh;
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Threshold threshold = (Threshold) o;
+        return getThresholdTarget() == threshold.getThresholdTarget();
```
christ66: Since you're using `Objects.hash` below for the hashcode it would be best to use `Objects.equal` in here instead of using the `==` operator.

PR #10, opened 2018-06-05
[JENKINS-51611] Add several tests

cizezsy:
- [x] If there is no coverage report file.
- [x] If there is a coverage file, but it's empty.
- [ ] A large coverage file (several hundred MB in size. Would need to probably zip the file).


cizezsy:
@christ66 OK, I will create a ticket for it as a future task.

PR #14, opened 2018-06-07
Create LICENSE

cizezsy:
Is this license like this OK?


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 Jenkins
```
oleg-nenashev: `Shenyu Zheng and other Jenkins contributors` ?

PR #16, opened 2018-06-08
[JENKINS-51737] Move auto-detector coverage publishing logic to the e…

cizezsy:
…xtension point implementation

### DIFF
--- a/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
+++ b/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
@@ -12,6 +12,7 @@ import hudson.remoting.VirtualChannel;
 import io.jenkins.plugins.coverage.adapter.CoverageReportAdapter;
 import io.jenkins.plugins.coverage.adapter.CoverageReportAdapterDescriptor;
 import io.jenkins.plugins.coverage.adapter.Detectable;
+import io.jenkins.plugins.coverage.detector.Detector;
 import io.jenkins.plugins.coverage.exception.CoverageException;
 import io.jenkins.plugins.coverage.targets.CoverageElement;
 import io.jenkins.plugins.coverage.targets.CoverageResult;
@@ -19,8 +20,6 @@ import io.jenkins.plugins.coverage.targets.Ratio;
 import io.jenkins.plugins.coverage.threshold.Threshold;
 import jenkins.MasterToSlaveFileCallable;
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.io.IOUtils;
-import org.apache.commons.lang.StringUtils;
 import org.jvnet.localizer.Localizable;
 
 import javax.annotation.Nonnull;
@@ -34,7 +33,6 @@ import java.io.PrintStream;
 import java.lang.reflect.Constructor;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Iterator;
@@ -58,7 +56,7 @@ public class CoverageProcessor {
     private boolean failUnstable;
     private boolean failNoReports;
 
-    private String autoDetectPath;
+    private Detector reportDetector;
 
 
     /**
@@ -141,10 +139,11 @@ public class CoverageProcessor {
             }
         }
 
-        // If enable automatically detecting, it will try to find report and correspond adapter.
-        if (isEnableAutoDetect()) {
-            logger.println("Auto Detect is enabled: Looking for reports...");
-            List<FilePath> detectedFilePaths = Arrays.stream(workspace.act(new FindReportCallable(autoDetectPath, null)))
+        // if have report detector, start detect report and its correspond adapter
+        if (getReportDetector() != null) {
+            logger.println("Detector is enabled: Looking for reports automatically...");
+            List<FilePath> detectedFilePaths = reportDetector.findFiles(workspace)
+                    .stream()
                     .filter(filePath -> {
                         for (Map.Entry<CoverageReportAdapter, Set<FilePath>> entry : reports.entrySet()) {
                             if (entry.getValue().contains(filePath))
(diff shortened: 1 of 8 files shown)


src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
```diff
@@ -40,11 +41,11 @@
     private boolean failUnstable;
     private boolean failNoReports;
 
-    private String autoDetectPath;
+    private Detector reportDetector;
 
     @DataBoundConstructor
-    public CoveragePublisher(String autoDetectPath) {
-        this.autoDetectPath = autoDetectPath;
+    public CoveragePublisher(Detector reportDetector) {
```
oleg-nenashev: Is it really required? You already have DataBoundSetter

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -388,19 +387,21 @@ private CoverageResult aggregatedResults(Map<CoverageReportAdapter, List<Coverag
...
 
     /**
-     * @return <code>true</code> if auto detect is enable
+     * Getter for property 'reportDetector'.
+     *
+     * @return value for property 'reportDetector'
      */
-    public boolean isEnableAutoDetect() {
-        return !StringUtils.isEmpty(autoDetectPath);
+    public Detector getReportDetector() {
```
oleg-nenashev: `@CheckForNull` would help

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -388,19 +387,21 @@ private CoverageResult aggregatedResults(Map<CoverageReportAdapter, List<Coverag
...
      */
-    public void setAutoDetectPath(String autoDetectPath) {
-        this.autoDetectPath = autoDetectPath;
+    public void setReportDetector(Detector reportDetector) {
+        this.reportDetector = reportDetector;
     }
 
     /**
-     * @return <code>true</code> if auto detect is enable
+     * Getter for property 'reportDetector'.
```
oleg-nenashev: This comment does not hell much. Maybe it makes sense to just remove it
In such cases it's also better to use field links like `{@link #reportDetector}` so that field renaming gets applied in comments as well

src/main/resources/io/jenkins/plugins/coverage/CoveragePublisher/config.jelly
```diff
@@ -2,9 +2,7 @@
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
 
     <f:block>
-        <f:entry title="Ant-Style Path for Automatically Detect" field="autoDetectPath">
-            <f:textbox default="*.xml" value="${instance.autoDetectPath}"/>
-        </f:entry>
+        <f:dropdownDescriptorSelector field="reportDetector" title="Report Detector" descriptors="${descriptor.listDetectorDescriptors}" />
```
oleg-nenashev: IIUC there should be another noop extension point implementation in such case

cizezsy: noop extension? I don't understand....

src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
cizezsy: I think `detector` is a necessary parameter, so I add it to the constructor.

src/main/resources/io/jenkins/plugins/coverage/CoveragePublisher/config.jelly
cizezsy: Oh, I know. You mean I should implement a noop detector?

src/main/java/io/jenkins/plugins/coverage/detector/Detector.java
```diff
@@ -0,0 +1,21 @@
...
+import hudson.model.Describable;
+import hudson.model.Descriptor;
+import jenkins.model.Jenkins;
+
+import java.io.IOException;
+import java.util.List;
+
+public abstract class Detector implements ExtensionPoint, Describable<Detector> {
+
+    public abstract List<FilePath> findFiles(FilePath workspace) throws IOException, InterruptedException;
```
oleg-nenashev: It should have `TaskListener` at least. And maybe `Run` object

src/main/java/io/jenkins/plugins/coverage/detector/DetectorDescriptor.java
```diff
@@ -0,0 +1,30 @@
...
+        this.detectorName = detectorName;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Nonnull
+    @Override
+    public String getDisplayName() {
+        return detectorName;
```
oleg-nenashev: This is a bad approach, because you won't be able to make strings localizable

src/main/java/io/jenkins/plugins/coverage/detector/DetectorDescriptor.java
```diff
@@ -0,0 +1,30 @@
...
+
+import hudson.DescriptorExtensionList;
+import hudson.model.Descriptor;
+import jenkins.model.Jenkins;
+
+import javax.annotation.Nonnull;
+
+public class DetectorDescriptor<T extends Detector> extends Descriptor<Detector> {
+
+    private String detectorName;
```
oleg-nenashev: Just remove it for now?

src/main/java/io/jenkins/plugins/coverage/detector/AntPathDetector.java
```diff
@@ -0,0 +1,55 @@
...
+        return Arrays.stream(workspace.act(new MasterToSlaveFileCallable<FilePath[]>() {
+            @Override
+            public FilePath[] invoke(File f, VirtualChannel channel) throws IOException, InterruptedException {
+                return workspace.list(path);
+            }
+        })).collect(Collectors.toList());
+    }
+
+
+    @Symbol("antDetector")
```
oleg-nenashev: `antPath()` ? Another option would be `wildcard`, but it's not fully correct

cizezsy:
My implementation in this pull request is bad, so I will close this one and open a new pull request for JENKINS-51737.

PR #22, opened 2018-06-20
[JENKINS-51737] Move auto-detector coverage publishing logic to the e…

cizezsy:
…xtension point implementation

- replace CoverageReportAdapter extension point with CoverageAdapter extension point
- move detector logic to extension point
- other trivial fixs


cizezsy:
I have tried to break pull request into smaller pieces but failed...So how to make a pull request smaller in this case?

PR #23, opened 2018-06-22
[JENKINS-51988] Add source code navigation.

cizezsy:


src/main/webapp/css/style.css
```diff
@@ -57,62 +57,7 @@ table.source tr.noCover {
...
-    border: 0px;
-    margin: 0px;
-    padding: 0px;
-    padding-left: 4px;
-}
-
-table.percentgraph td.data {
-  text-align: right;
-}
+Wx
```
oleg-nenashev: hm?

src/main/java/io/jenkins/plugins/coverage/source/SourceFileResolver.java
```diff
@@ -0,0 +1,192 @@
...
+
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.util.Map;
+
+public class SourceFileResolver implements ExtensionPoint, Describable<SourceFileResolver> {
```
oleg-nenashev: Seems the implementation is heavily integrated into the extension point. I would suggest splitting it to the implementation and abstract class

src/main/java/io/jenkins/plugins/coverage/source/DefaultSourceFileResolver.java
```diff
@@ -0,0 +1,146 @@
...
+
+            paint.setTotalLines(line);
+        } catch (IOException | InterruptedException e) {
+            throw new CoverageException(e);
+        }
+    }
+
+
+
+    @Extension
```
oleg-nenashev: Symbol is missing, needed for Pipeline integration

PR #26, opened 2018-07-09
[JENKINS-51736] Add support of llvm-cov Report converter

cizezsy:
- [x] implement llvm-cov JSON format report adapter 
- [x] write java document comment
- [x] clean code


src/main/java/io/jenkins/plugins/coverage/adapter/LLVMCovReportAdapter.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+
+    public static class LLVMCovDocumentConverter extends JSONDocumentConverter {
+
+        @Override
+        protected Document convert(JSONObject report, Document document) throws CoverageException {
+            // only support 2.0.0 version now
+            if (!report.getString("version").equals("2.0.0") || !report.getString("type").equals("llvm.coverage.json.export")) {
+                throw new CoverageException("Unsupported Json file");
```
jeffpearce: Could we separate these two checks, and provide more detailed exceptions, e.g. `"Unsupported Json file - type must be "llvm.coverage.json.export"`? This would help the user better understand what's wrong

cizezsy: Yeah

pom.xml
```diff
@@ -84,6 +84,11 @@
             <version>3.0.3</version>
             <type>jar</type>
         </dependency>
+        <dependency>
+            <groupId>com.alibaba</groupId>
+            <artifactId>fastjson</artifactId>
+            <version>1.2.47</version>
```
oleg-nenashev: i highly recommend using http://plugins.jenkins.io/jackson2-api unless there is a specific need to use a custom JSON library

cizezsy: OK, I will change it.

cizezsy:
@oleg-nenashev Do you think is pull request is fine? I want to merge this pull request, then divide the llvm part to an individual plugin.

src/test/java/io/jenkins/plugins/coverage/CoveragePublisherPipelineTest.java
```diff
@@ -47,26 +48,30 @@ public void testAutoDetect() throws Exception {
     @Test
     public void testAdapters() throws Exception {
         CoverageScriptedPipelineScriptBuilder builder = CoverageScriptedPipelineScriptBuilder.builder()
-                .addAdapter(new CoberturaReportAdapter("cobertura-coverage.xml"))
-                .addAdapter(new JacocoReportAdapter("jacoco.xml"));
+//                .addAdapter(new CoberturaReportAdapter("cobertura-coverage.xml"))
+//                .addAdapter(new JacocoReportAdapter("jacoco.xml"));
```
jeffpearce: Can we remove the commented code?

src/test/java/io/jenkins/plugins/coverage/CoveragePublisherPipelineTest.java
```diff
@@ -47,26 +48,30 @@ public void testAutoDetect() throws Exception {
...
-                .copyFrom(getClass().getResourceAsStream("cobertura-coverage.xml"));
-        workspace.child("jacoco.xml").copyFrom(getClass()
-                .getResourceAsStream("jacoco.xml"));
+                .child("llvm-cov.json")
+                .copyFrom(getClass().getResourceAsStream("llvm-cov.json"));
+//        Objects.requireNonNull(workspace)
+//                .child("cobertura-coverage.xml")
+//                .copyFrom(getClass().getResourceAsStream("cobertura-coverage.xml"));
+//        workspace.child("jacoco.xml").copyFrom(getClass()
+//                .getResourceAsStream("jacoco.xml"));
```
jeffpearce: Can we remove this?

src/test/java/io/jenkins/plugins/coverage/CoveragePublisherPipelineTest.java
cizezsy: OK, I forgot about it :P

PR #27, opened 2018-07-20
[JENKINS-52628] Change current REST API to the standard format.

cizezsy:
Before we get data by REST API like this: `api/trend/json`
Now we get data by REST API like this: `trend/api/json`


src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
```diff
@@ -502,6 +503,17 @@ public CoverageResult getPreviousResult() {
...
     public Object getDynamic(String token, StaplerRequest req, StaplerResponse rsp) throws IOException {
         token = token.toLowerCase();
+        String restPath = req.getRestOfPath();
+
+        // prevent name conflict
+        if (restPath.startsWith("/api/")) {
+            if (token.equals("trend")) {
+                return new RestResultWrapper(new CoverageTrendTree(getName(), getCoverageTrends(), getChildrenReal()));
+            } else if (token.equals("result")) {
+                return new RestResultWrapper(this);
```
oleg-nenashev: why do you need a wrapper here? it seems yo introduce recursive API

cizezsy: @oleg-nenashev If I return `new Api(this)` directly, we need to retrive data by URL like
 `.../result/json` instead of `.../result/api/json`. 
And I do not think it will cause a recursive call, one token map one object...

PR #37, opened 2018-08-03
[JENKINS-51733], [JENKINS-52839] - Refactoring of the adopted code, M…

cizezsy:
…ake Code Coverage API plugin more generic


src/main/java/io/jenkins/plugins/coverage/adapter/JavaCoverageReportAdapterDescriptor.java
```diff
@@ -0,0 +1,22 @@
...
+public class JavaCoverageReportAdapterDescriptor extends CoverageReportAdapterDescriptor<CoverageReportAdapter> {
+
+    public JavaCoverageReportAdapterDescriptor(Class<? extends CoverageReportAdapter> clazz) {
+        super(clazz);
+    }
+
+    @Override
+    public List<CoverageElement> getCoverageElements() {
+        return Arrays.asList(new CoverageElement("Java Group", 0),
+                new CoverageElement("Java Package", 1),
```
oleg-nenashev: It may be a proble. `Cobertura` plugin and other such plugin are widely used to report coverage of non-Java code, even with code browsing.  I would advice to have a way to specify the code type in the implementations

src/test/java/io/jenkins/plugins/coverage/adapter/CoberturaReportAdapter.java
```diff
@@ -40,8 +40,7 @@ public String getXSD() {
 
     @Symbol("cobertura")
     @Extension
-    public static final class CoverturaReportAdapterDescriptor
-            extends CoverageReportAdapterDescriptor<CoberturaReportAdapter>
+    public static final class CoverturaReportAdapterDescriptor extends JavaCoverageReportAdapterDescriptor
```
oleg-nenashev: Typo. Cobertura, not Covertura

PR #39, opened 2018-08-07
[Fix] Fix a bug that methods have same name will be view as same elem…

cizezsy:
…ent.

Add method signature to differentiate the methods have the same name in the same class.


oleg-nenashev:
Oh right. If this is a part of the spec, I am fine with that. Maybe makes sense to explicitly mention it in Javadoc

PR #40, opened 2018-08-07
[JENKINS-51926] Supporting combining reports within a build(e.g. afte…

cizezsy:
…r parallel() execution in Pipeline)

### DIFF
--- a/src/main/java/io/jenkins/plugins/coverage/CoverageAction.java
+++ b/src/main/java/io/jenkins/plugins/coverage/CoverageAction.java
@@ -21,6 +21,9 @@ public class CoverageAction implements StaplerProxy, SimpleBuildStep.LastBuildAc
     private transient WeakReference<CoverageResult> report;
     private HealthReport healthReport;
 
+    private String cacheBranchName;
+    private int branchCount = 1;
+
     public CoverageAction(CoverageResult result) {
         this.report = new WeakReference<>(result);
     }
@@ -145,4 +148,19 @@ public class CoverageAction implements StaplerProxy, SimpleBuildStep.LastBuildAc
     }
 
 
+    public String getCacheBranchName() {
+        return cacheBranchName;
+    }
+
+    public void setCacheBranchName(String cacheBranchName) {
+        this.cacheBranchName = cacheBranchName;
+    }
+
+    public int getBranchCount() {
+        return branchCount;
+    }
+
+    public void setBranchCount(int branchCount) {
+        this.branchCount = branchCount;
+    }
 }
--- a/src/main/java/io/jenkins/plugins/coverage/CoverageAction.java
+++ b/src/main/java/io/jenkins/plugins/coverage/CoverageAction.java
@@ -20,7 +20,6 @@ public class CoverageAction implements StaplerProxy, SimpleBuildStep.LastBuildAc
     private transient Run<?, ?> owner;
     private transient WeakReference<CoverageResult> report;
     private HealthReport healthReport;
-    private int callTimes;
 
     private String cacheBranchName;
     private int branchCount = 1;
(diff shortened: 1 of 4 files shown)


src/main/java/io/jenkins/plugins/coverage/CoverageAction.java
```diff
@@ -21,6 +21,9 @@
     private transient WeakReference<CoverageResult> report;
     private HealthReport healthReport;
 
+    private String cacheBranchName;
+    private int branchCount = 1;
```
oleg-nenashev: I am not sure "count" is the best term for it. CC @jeffpearce

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -315,6 +314,57 @@ private HealthReport processThresholds(Map<CoverageReportAdapter, List<CoverageR
         return new HealthReport(score, localizeDescription);
     }
 
+
+    private void covertResultsToAction(CoverageResult coverageReport, HealthReport healthReport) throws IOException {
```
oleg-nenashev: Typo: "coNvert"

src/main/java/io/jenkins/plugins/coverage/CoverageAction.java
jeffpearce: How about `branchNumber` or `branchIndex`?

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -315,6 +312,57 @@ private HealthReport processThresholds(Map<CoverageReportAdapter, List<CoverageR
         return new HealthReport(score, localizeDescription);
     }
 
+
+    private void convertResultsToAction(CoverageResult coverageReport, HealthReport healthReport) throws IOException {
+        synchronized (CoverageProcessor.class) {
+            CoverageAction previousAction = run.getAction(CoverageAction.class);
+
+            if (previousAction != null) {
```
jeffpearce: What should happen if previousAction is null?

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -315,6 +312,57 @@ private HealthReport processThresholds(Map<CoverageReportAdapter, List<CoverageR
...
+                CoverageResult previousResult = previousAction.getResult();
+
+                int branchCount = previousAction.getBranchCount();
+                if (branchCount == 1) {
+                    String branchNameOfFirstCall = previousAction.getCacheBranchName();
```
jeffpearce: I'm not sure about the word `cache` here - is it really a cached? Maybe just `getBranchName`?

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
cizezsy: If `previousAction` is null, it means this is the first time plugin be called at the current build. It will aggregate current reports and save them, create action and save the first branch name to action.

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
cizezsy: This branch name is the name of the branch to be first called. So it actually saves the first branch name. But I agree with your, `cache` is not very good.

src/main/java/io/jenkins/plugins/coverage/CoverageAction.java
cizezsy: I think `branchIndex` is good.

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
cizezsy: Maybe I should change it to `branchNameOfFirstCall`?

cizezsy:
I think the previous implementation is bad, so I rewrite them all.

PR #41, opened 2018-08-12
[Fix] Fix a bug cause Jacoco show incorrect coverage, other trivial fix, update version prepare for 1.0 releasing

cizezsy:


pom.xml
```diff
@@ -9,7 +9,7 @@
     </parent>
     <groupId>io.jenkins.plugins</groupId>
     <artifactId>code-coverage-api</artifactId>
-    <version>1.0.0-SNAPSHOT</version>
+    <version>1.0.0</version>
```
oleg-nenashev: I doubt it's what you want. Maven release plugin does it automatically

src/main/java/io/jenkins/plugins/coverage/targets/Ratio.java
```diff
@@ -150,18 +150,18 @@ public int hashCode() {
...
-        if (xx == x && yy == y) {
-            int idx = yy * (yy + 1) / 2 + xx;
-            if (0 <= idx && idx < COMMON_INSTANCES.length) {
-                Ratio r = COMMON_INSTANCES[idx];
-                if (r == null)
-                    COMMON_INSTANCES[idx] = r = new Ratio(x, y);
-                return r;
-            }
-        }
+//        int xx = (int) x;
```
oleg-nenashev: the code should be removed OR there should be a comment explaining why it's commented out

pom.xml
cizezsy: Yeah, you are right, I forget that.

cizezsy:
@oleg-nenashev Sorry, I will create a ticket for such case in the future.

src/main/resources/io/jenkins/plugins/coverage/targets/CoverageResult/index.jelly
```diff
@@ -62,7 +61,7 @@
                             coverageChartGenerator.generateSummaryChart(results, 'summaryChart', name)
                         });
 
-                        var element = '${it.childElement.displayName}';
+                        var element = '${it.childElement.name}';
```
oleg-nenashev: This does not like to be correct. Localized display names should be used in UI

cizezsy: This part is adopted from Cobertura plugin before. But now, it is the name of `CoverageElement` and registered by others and we need to it as the key to save and restore coverage. So it cannot be Localized.

oleg-nenashev: It needs a follow-up JIRA task at least. There should be a way to localize the report layouts, but it can be done later

cizezsy: OK, I have created one.

PR #42, opened 2018-08-14
[JENKINS-53023] - Fix broken coverage calculating

cizezsy:
- fix incorrect `conditionals` calculating for Jacoco.
- ratio `0/0` will be regard as no data before, now it will be view as missed.
- improve result page.


cizezsy:
I think I can release 1.0, once this pull request is merged.

PR #83, opened 2019-05-09
Show diff in coverage for change request builds

tempora-mutantur:
It would be very convenient to see the code coverage diff right in Jenkins. If our
build is going for a multibranch change request job  - we take the
target branch commit that is used for our change request build. Then, we search through the
last fixed amount of builds the build that was built for this target branch commit. If
we find it - we check if it has line code coverage information. In case it has it -  we calculate diff between the current code coverage and this target branch build coverage.

Examples:
Increased code coverage:
image

Decreased code coverage:
image

No changes:
image

Thank you.

### DIFF
--- a/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
+++ b/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
@@ -2,23 +2,31 @@ package io.jenkins.plugins.coverage;
 
 
 import com.google.common.collect.Sets;
+
 import hudson.FilePath;
 import hudson.model.HealthReport;
+import hudson.model.Job;
+import hudson.model.ItemGroup;
 import hudson.model.Result;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.remoting.VirtualChannel;
+import hudson.util.RunList;
 import io.jenkins.plugins.coverage.adapter.CoverageReportAdapter;
 import io.jenkins.plugins.coverage.adapter.CoverageReportAdapterDescriptor;
 import io.jenkins.plugins.coverage.detector.Detectable;
 import io.jenkins.plugins.coverage.detector.ReportDetector;
 import io.jenkins.plugins.coverage.exception.CoverageException;
 import io.jenkins.plugins.coverage.source.SourceFileResolver;
-import io.jenkins.plugins.coverage.targets.CoverageElement;
-import io.jenkins.plugins.coverage.targets.CoverageResult;
-import io.jenkins.plugins.coverage.targets.Ratio;
+import io.jenkins.plugins.coverage.targets.*;
 import io.jenkins.plugins.coverage.threshold.Threshold;
+import jenkins.branch.MultiBranchProject;
 import jenkins.MasterToSlaveFileCallable;
+import jenkins.scm.api.SCMHead;
+import jenkins.scm.api.SCMRevision;
+import jenkins.scm.api.SCMRevisionAction;
+import jenkins.scm.api.mixin.ChangeRequestSCMHead;
+import jenkins.scm.api.mixin.ChangeRequestSCMRevision;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 import org.jvnet.localizer.Localizable;
@@ -45,6 +53,8 @@ public class CoverageProcessor {
     private boolean failNoReports;
 
     private String globalTag;
+
+    private boolean calculateDiffForChangeRequests;
 
     private SourceFileResolver sourceFileResolver;
 
@@ -96,7 +106,9 @@ public class CoverageProcessor {
 
         HealthReport healthReport = processThresholds(results, globalThresholds);
 
-
+        if (calculateDiffForChangeRequests) {
+            setDiffInCoverageForChangeRequest(coverageReport);
+        }
         convertResultToAction(coverageReport, healthReport);
     }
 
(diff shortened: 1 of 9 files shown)


src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -143,6 +152,65 @@ private void convertResultToAction(CoverageResult coverageReport, HealthReport h
         }
     }
 
+    private void setDiffInCoverageForChangeRequest(CoverageResult coverageReport) {
+        // Use limited number of builds to don't search too long
+        final int numberOfBuildsFromTargetBranch = 50;
+        // Calculate diff only for line coverage
+        Ratio changeRequestLinesCoverage = coverageReport.getCoverage(CoverageElement.LINE);
+        if (changeRequestLinesCoverage != null) {
```
cizezsy: Could you reduce the nesting level? There are too many nested if statement and hard to recognize.

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -97,6 +105,7 @@ public void performCoverageReport(List<CoverageReportAdapter> reportAdapters, Li
         HealthReport healthReport = processThresholds(results, globalThresholds);
 
 
+        setDiffInCoverageForChangeRequest(coverageReport);
```
cizezsy: Should we make it a configurable option?

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
tempora-mutantur: Agree, will try

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
tempora-mutantur: Yes, I think it would be better to make it configurable. Will change it.

cizezsy: @tandibar I think we should config it in Publisher level :)

tempora-mutantur: @cizezsy Please, check

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
tempora-mutantur: @cizezsy Please, check

cizezsy: @tempora-mutantur Maybe we can use some codes like this
```
if (changeRequestLinesCoverage == null) {
    return
}

Job job = run.getParent();
parent = job.getParent();

if (!(parent instanceof MultiBranchProject)) {
    return
}

......
```

Your current approach will pass through all if statement even if the condition must be false.

src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
```diff
@@ -165,6 +168,15 @@ public void setTag(String tag) {
         this.tag = tag;
     }
 
+    public boolean getCalculateDiffForChangeRequests() {
+        return calculateDiffForChangeRequests;
+    }
+
+    @DataBoundSetter
+    public void setCalculateDiffForChangeRequests(boolean calculateDiffForChangeRequests) {
```
cizezsy: We need to add a checkbox in CoveragePublisher/config.jelly to config it.

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
tempora-mutantur: @cizezsy 
Thank you. Please, take a look, I've made such change

TinyZzh:
is this just diff numeric with two branch ?
 
why not we analysis the total detail diff coverage with each file?  
for example: usually a PR includes some git commits.  and we want to known, how much code coverage increase or decrease by this PR. just with this commits.    

@cizezsy @tempora-mutantur

tempora-mutantur:
> why not we analysis the total detail diff coverage with each file?

Sorry, I'm not sure what you mean. This calculates diff between your PR **build** and the target branch **build** that was built on a commit that your PR is based on. This final number, the diff, is the total line coverage diff, one of the most important things that tools like codecov and coveralls do. So you can understand if your PR reduces/increases code coverage. If you build every commit of your PR you'll have the statistics for every commit.

Diff for each file can take a lot of time, I mean, machine time, if I understand this correctly. We may add this functionality further, but this should be optimised really hard. Tools like codecov and coveralls do this asynchronously.

TinyZzh:
sorry for my poor English.  
my company have a large and old java project.  we want calculate PR's changed code's coverage without the previous code. a little like codecov's relative coverage feature.  

found the codecov's docs.  use git diff command to calculates relative code coverage. docs link:
 [CodeCov's feature comparing-commits](https://docs.codecov.io/docs/comparing-commits) .

tempora-mutantur:
@TinyZzh Sorry, I don't understand what you mean by "without the previous code". Actually, you can think about this that this is a diff between two commits, one from the target branch and the other from is the PR head.

tempora-mutantur:
@cizezsy Hi. Are there any other problems with this PR? Maybe we can think of them together?

src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
```diff
@@ -49,6 +49,8 @@
 
     private String tag;
 
+    private boolean calculateDiffForChangeRequests = true;
```
cizezsy: Maybe we should set the default value to false.

src/main/resources/io/jenkins/plugins/coverage/CoveragePublisher/config.jelly
```diff
@@ -19,6 +19,9 @@
             <div>
                 <f:checkbox field="failNoReports" title="Failed if No Reports Found"/>
             </div>
+            <div>
+                <f:checkbox field="calculateDiffForChangeRequests" title="If it is a change request build, calculate code coverage diff with a target branch build" checked="true"/>
```
cizezsy: Same as the above.

src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
tempora-mutantur: Agree, thanks

src/main/resources/io/jenkins/plugins/coverage/CoveragePublisher/config.jelly
tempora-mutantur: Done

cizezsy:
@tempora-mutantur Thanks for your contribution! merged it~

Issue #82, opened 2019-05-08
Source files partially copied when transient agent is destroyed (i.e. Docker) 


capadgett:
Thanks for the awesome plugin (efficient, snappy interface, etc.).

When using a transient docker agent, I see files in the build directory on the Jenkins master that are partially copied or empty, and have the trailing `_copied` at the end of the file name.

I am not positive, but I think this issue is related to not waiting for the ExecutorService in `io/jenkins/plugins/coverage/source/DefaultSourceFileResolver.java`:
```
   81         ExecutorService service = Executors.newFixedThreadPool(5);
   82
   83         paints.forEach((sourceFilePath, paint) -> {
   84             service.submit(() -> {                                              
```

I should also mention that I am using the `publishCoverage` pipeline step in a `stage{ post{} }`, not a `pipeline{ post{} }`, but I wouldn't think that should matter.

Might be related to fix for #77

cizezsy:
@capadgett You are right, ExecutorService cannot prevent agent from destroyed so not be able to copy all source files. Thanks for the feedback, I will try to fix it in next release.

cizezsy:
@capadgett Fixed in version 1.0.11, thanks :P

PR #86, opened 2019-05-13
Use CountDownLatch to wait all copy taks completed.

cizezsy:
Source files will partially copied when using transient agent.

see: https://github.com/jenkinsci/code-coverage-api-plugin/issues/82

### DIFF
--- a/src/main/java/io/jenkins/plugins/coverage/source/DefaultSourceFileResolver.java
+++ b/src/main/java/io/jenkins/plugins/coverage/source/DefaultSourceFileResolver.java
@@ -40,8 +40,10 @@ import org.kohsuke.stapler.DataBoundConstructor;
 import java.io.*;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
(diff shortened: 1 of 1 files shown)


src/main/java/io/jenkins/plugins/coverage/source/DefaultSourceFileResolver.java
```diff
@@ -115,9 +119,19 @@ public void resolveSourceFiles(Run<?, ?> run, FilePath workspace, TaskListener l
...
-        });
+            } finally {
+                // ensure latch will count down
+                latch.countDown();
+            }
+        }));
+
 
+        try {
+            latch.await(1, TimeUnit.HOURS);
```
capadgett: Maybe add to the list of future improvements to make this wait time a user configurable setting.

cizezsy: Good idea, and will add it in the future.

PR #88, opened 2019-05-14
[feature] calculate newly code's coverage

TinyZzh:
Calculate  the code coverage of newly modified code.

# the solution
the solution.  reference [Codecov docs](https://docs.codecov.io/docs/comparing-commits)
1. use git 's command "git diff" to calculate changed source code blocks.
2. calculate newly code's coverage by step one's data and coverage report.  
3. newly code's coverage = covered total lines /  （ covered total  lines +  uncovered total lines）

# examples
1. source code file not change
incalculable.  can not calculate the newly code coverage.

2. source code file changed
compare trigger jenkins's build action's git commit  with the previous jenkins no failed build's git commit。
this example is add a new class named Orange.java,  add 9 lines.  junit test coverd 6 lines..  the newly code coverage is 66.67%(= 6/9).  
image

@cizezsy Please, take a look


TinyZzh:
image

TinyZzh:
fix some bugs and make the vcs setting easy

cizezsy:
@TinyZzh 
Thanks very much for your contribution, this is a really nice feature and your implementation is also good.

But, I am not sure if we should add this to plugin.... :( This implementation is dependent on Git.
If we use Mercurial, Subversion and other VCS, we can't use this feature. Although we can add support for other VCS, this will make the plugin more complex. 

Also, this plugin will only get data from coverage reports and previous builds now, so it makes the structure clearer and tool-independent. This implementation will read data from .git and it seems to do the things that **coverage tools** should do. We also need to store .git to build directory if we use transigent agent.

I think it is better to create a new plugin that gets coverage data from code-coverage-api-plugin and analysis from local VCS..

TinyZzh:
emmm...

PR #119, opened 2019-09-06
Feat: add release drafter

cizezsy:


oleg-nenashev:
The config is for 2-digit versions. Maybe you want to have a 3-digit config like in ci.jenkins.io-runner

cizezsy:
Oh, you are right, fixed it now, thanks~

PR #130, opened 2019-11-12
Avoid thread leak by copying files synchronously

recampbell:
We see unbounded thread growth with the following stack trace:

```
java.lang.Throwable: created pool #47
at java.util.concurrent.Executors$DefaultThreadFactory.<init>(Executors.java:610)
at java.util.concurrent.Executors.defaultThreadFactory(Executors.java:353)
at java.util.concurrent.ThreadPoolExecutor.<init>(ThreadPoolExecutor.java:1203)
at java.util.concurrent.Executors.newFixedThreadPool(Executors.java:89)
at io.jenkins.plugins.coverage.source.DefaultSourceFileResolver.resolveSourceFiles(DefaultSourceFileResolver.java:83)
at io.jenkins.plugins.coverage.CoverageProcessor.performCoverageReport(CoverageProcessor.java:115)
at io.jenkins.plugins.coverage.CoveragePublisher.perform(CoveragePublisher.java:94)
at org.jenkinsci.plugins.workflow.steps.CoreStep$Execution.run(CoreStep.java:80)
at org.jenkinsci.plugins.workflow.steps.CoreStep$Execution.run(CoreStep.java:67)
at org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution.lambda$start$0(SynchronousNonBlockingStepExecution.java:47)
at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
at java.util.concurrent.FutureTask.run(FutureTask.java:266)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
at java.lang.Thread.run(Thread.java:748)
```

Replicates what @jglick  did in jenkinsci/extreme-notification-plugin#8

### DIFF
--- a/src/main/java/io/jenkins/plugins/coverage/source/DefaultSourceFileResolver.java
+++ b/src/main/java/io/jenkins/plugins/coverage/source/DefaultSourceFileResolver.java
@@ -32,6 +32,7 @@ import io.jenkins.plugins.coverage.BuildUtils;
 import io.jenkins.plugins.coverage.exception.CoverageException;
 import io.jenkins.plugins.coverage.targets.CoveragePaint;
 import jenkins.MasterToSlaveFileCallable;
+import jenkins.util.Timer;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.exception.ExceptionUtils;
 import org.jenkinsci.Symbol;
@@ -41,8 +42,6 @@ import java.io.*;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 public class DefaultSourceFileResolver extends SourceFileResolver {
@@ -80,11 +79,10 @@ public class DefaultSourceFileResolver extends SourceFileResolver {
 
         listener.getLogger().printf("%d source files need to be copied%n", paints.size());
 
-        ExecutorService service = Executors.newFixedThreadPool(5);
         // wait until all tasks completed
         CountDownLatch latch = new CountDownLatch(paints.entrySet().size());
 
-        paints.forEach((sourceFilePath, paint) -> service.submit(() -> {
+        paints.forEach((sourceFilePath, paint) -> Timer.get().submit(() -> {
             try {
                 FilePath[] possibleFiles;
                 try {
--- a/src/main/java/io/jenkins/plugins/coverage/source/DefaultSourceFileResolver.java
+++ b/src/main/java/io/jenkins/plugins/coverage/source/DefaultSourceFileResolver.java
@@ -32,7 +32,6 @@ import io.jenkins.plugins.coverage.BuildUtils;
 import io.jenkins.plugins.coverage.exception.CoverageException;
 import io.jenkins.plugins.coverage.targets.CoveragePaint;
 import jenkins.MasterToSlaveFileCallable;
-import jenkins.util.Timer;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.exception.ExceptionUtils;
 import org.jenkinsci.Symbol;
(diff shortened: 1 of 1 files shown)


recampbell:
cc: @varyvol

src/main/java/io/jenkins/plugins/coverage/source/DefaultSourceFileResolver.java
```diff
@@ -80,11 +79,10 @@ public void resolveSourceFiles(Run<?, ?> run, FilePath workspace, TaskListener l
 
         listener.getLogger().printf("%d source files need to be copied%n", paints.size());
 
-        ExecutorService service = Executors.newFixedThreadPool(5);
         // wait until all tasks completed
         CountDownLatch latch = new CountDownLatch(paints.entrySet().size());
 
-        paints.forEach((sourceFilePath, paint) -> service.submit(() -> {
+        paints.forEach((sourceFilePath, paint) -> Timer.get().submit(() -> {
```
varyvol: Given this is inside a loop, I'm not sure this could affect performance and it would be better to fix the previous implementation by:
- Shutting down the executor.
- Adding a naming thread factory.

@jglick what's your opinion here?

jglick: Without knowing a typical value for `paints.size()`, or how long each task runs, it is hard to say. (`Timer` is a global pool of 10 threads, intended for quick tasks.) Also, what is calling `resolveSourceFiles`—a build thread? From a brief glance at the code, I would think it would be better to just run these tasks serially, since they probably spend most time in `FilePath.copyTo(FilePath)` on the same agent, which cannot be usefully parallelized anyway. Would need some context from @cizezsy here.

src/main/java/io/jenkins/plugins/coverage/source/DefaultSourceFileResolver.java
```diff
@@ -80,11 +79,10 @@ public void resolveSourceFiles(Run<?, ?> run, FilePath workspace, TaskListener l
 
         listener.getLogger().printf("%d source files need to be copied%n", paints.size());
 
-        ExecutorService service = Executors.newFixedThreadPool(5);
```
alexanderrtaylor: What if instead of relying on the timer threads, we just put a timeout or add the shutdown in the finally block?

src/main/java/io/jenkins/plugins/coverage/source/DefaultSourceFileResolver.java
recampbell: I just pushed a commit to implement Jesse's suggestion of doing it serially.

jglick: I suppose you should edit the PR title accordingly.

cizezsy: The `paints.size()` is the number of the source files that need to be copied. It will copy source files from the agent, render them and save the result to master. The time spent most in `FilePath.copyTo()` which will copy source files from the agent and `paintSourceCode()` which will render the source file and save it to master.

As `FilePath.copyTo()` cannot be parallelized, maybe we need to run `paintSourceCode()` parallelly to increase the performance? @jglick

jglick: You could only see a speedup if the system were otherwise mostly idle, and you need to support build interruption. I doubt it is worth making the code more complicated for a speculative optimization, especially as build time is a low priority for performance compared to UI responsiveness.

cizezsy: Yeah, you are right...approved this pr, thanks.

batmat:
@cizezsy thanks for merging this! Could you please consider releasing this fix? As per @ryancampbell description, this is causing issues in actual Jenkins instances.
Thank you!

PR #131, opened 2019-11-22
Improve source file resolving

### LINKED ISSUE #90: "Cannot found source..." error with Cobertura Reports with Fully Qualified Filename property
Repro Steps:
Generate your Cobertura reports with [Coverlet](https://github.com/tonerdo/coverlet) using the --format cobertura flag
Validate that the `filename` property is filled with the fully qualified path of the file.
Send the report to the code-coverage-api and save the last build
Notice that the source cannot be found. 

Note: I will update the Repro steps with something more concrete this evening.

cizezsy:
- Fix #90.
- Rendering source file in agent.


tuky:
For us, this unfortunately stopped relative source file resolution from working. we're running

```
publishCoverage adapters: [istanbulCoberturaAdapter('coverage/coverage.xml')], ...
```

where the `<project root>/coverage/coverage.xml` will contain

```xml
<source>../src</source>
```

Downgrading to `1.1.1` fixes this for us again.

PR #152, opened 2020-04-16
Improved dropped coverage message

thatsIch:


tempora-mutantur:
Thank you

@cizezsy FYI

PR #169, opened 2020-07-20
Add CoverageChecksPublisher

XiongKezhi:
Example: https://github.com/XiongKezhi/codingstyle/pull/3/checks?check_run_id=888545163
- [x] Publish coverage as checks
- [x] Add tests
- [x] Add health report
- [x] Add configurations
- [ ] ...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -112,7 +112,7 @@
         <dependency>
             <groupId>io.jenkins.plugins</groupId>
             <artifactId>checks-api</artifactId>
-            <version>0.1.0-rc4.fc43ef9447c7</version>
+            <version>0.1.0</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -112,7 +112,7 @@
         <dependency>
             <groupId>io.jenkins.plugins</groupId>
             <artifactId>github-checks</artifactId>
-            <version>0.1.0-rc143.251559838646</version>
+            <version>0.1.0-rc148.efa7f48759e1</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -124,6 +124,18 @@
             <artifactId>spotbugs-annotations</artifactId>
             <version>4.0.2</version>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <version>3.4.4</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>3.16.1</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 
--- a/pom.xml
+++ b/pom.xml
@@ -111,8 +111,8 @@
         </dependency>
         <dependency>
             <groupId>io.jenkins.plugins</groupId>
-            <artifactId>github-checks</artifactId>
-            <version>0.1.0-rc148.efa7f48759e1</version>
+            <artifactId>checks-api</artifactId>
+            <version>0.1.0-rc4.fc43ef9447c7</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.0</version>
+        <version>4.3</version>
         <relativePath />
     </parent>
     <groupId>io.jenkins.plugins</groupId>
@@ -182,10 +182,6 @@
         <repository>
             <id>repo.jenkins-ci.org</id>
             <url>https://repo.jenkins-ci.org/public/</url>
-        </repository>
-        <repository>
-            <id>incrementals.jenkins-ci.org</id>
-            <url>https://repo.jenkins-ci.org/incrementals/</url>
         </repository>
     </repositories>
     <pluginRepositories>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.9</version>
+        <version>4.0</version>
         <relativePath />
     </parent>
     <groupId>io.jenkins.plugins</groupId>
@@ -13,18 +13,21 @@
     <packaging>hpi</packaging>
     <properties>
         <!-- Baseline Jenkins version you use to build the plugin. Users must have this version or newer to run. -->
-        <jenkins.version>2.60.3</jenkins.version>
+        <jenkins.version>2.204.4</jenkins.version>
         <java.level>8</java.level>
         <!-- Other properties you may want to use:
           ~ jenkins-test-harness.version: Jenkins Test Harness version you use to test the plugin. For Jenkins version >= 1.580.1 use JTH 2.0 or higher.
           ~ hpi-plugin.version: The HPI Maven Plugin version used by the plugin..
           ~ stapler-plugin.version: The Stapler Maven plugin version required by the plugin.
      -->
+
+        <!-- GitHub Checks plugin is still in beta version-->
+        <useBeta>true</useBeta>
     </properties>
     <name>Code Coverage API Plugin</name>
     <description>Code Coverage API Plugin</description>
     <url>https://github.com/jenkinsci/code-coverage-api-plugin</url>
-    
+
     <!-- The default licence for Jenkins OSS Plugins is MIT. Substitute for the applicable one if needed. -->
     <licenses>
         <license>
@@ -36,7 +39,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.9</version>
+            <version>1.20</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
@@ -104,7 +107,22 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>jackson2-api</artifactId>
-            <version>2.8.11.3</version>
+            <version>2.11.1</version>
+        </dependency>
+        <dependency>
+            <groupId>io.jenkins.plugins</groupId>
+            <artifactId>github-checks</artifactId>
+            <version>0.1.0-rc143.251559838646</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>display-url-api</artifactId>
+            <version>2.3.3</version>
+        </dependency>
+        <dependency>
+            <groupId>com.github.spotbugs</groupId>
+            <artifactId>spotbugs-annotations</artifactId>
+            <version>4.0.2</version>
         </dependency>
     </dependencies>
 
@@ -152,6 +170,10 @@
         <repository>
             <id>repo.jenkins-ci.org</id>
             <url>https://repo.jenkins-ci.org/public/</url>
+        </repository>
+        <repository>
+            <id>incrementals.jenkins-ci.org</id>
+            <url>https://repo.jenkins-ci.org/incrementals/</url>
         </repository>
     </repositories>
     <pluginRepositories>
--- a/src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
+++ b/src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
@@ -19,6 +19,8 @@ import java.util.*;
 class CoverageChecksPublisher {
     private final CoverageAction action;
     private final JenkinsFacade jenkinsFacade;
+    private final List<String> allTypes =
+            Arrays.asList("Report", "Group", "Package", "File", "Class", "Method", "Conditional", "Line", "Instruction");
 
     CoverageChecksPublisher(final CoverageAction action) {
         this(action, new JenkinsFacade());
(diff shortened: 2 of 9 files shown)


XiongKezhi:
cc @timja @uhafner @jeffpearce

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,104 @@
...
+
+    private ChecksDetails extractChecksDetails() {
+        CoverageResult result = action.getResult();
+        ChecksOutputBuilder outputBuilder = new ChecksOutputBuilder()
+                .withTitle("Code Coverage")
+                .withSummary("## Generated " + result.getResults().size() + " types of coverage.")
+                .withText(extractChecksText(result)); // TODO: improve summary
+
+        return new ChecksDetailsBuilder()
+                .withName("Code Coverage")
```
timja, line 36: not sure which field it is as they have the same text currently but can a summary go on the check status on the right hand side

image

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,104 @@
...
+        Map<CoverageElement, Ratio> ratios = result.getResults();
+        Map<String, Float> lastRatios = getLastRatios(result);
+
+        StringBuilder text = new StringBuilder();
+        for (Map.Entry<CoverageElement, Ratio> singleRatio : ratios.entrySet()) {
+            text.append("## ")
+                    .append(singleRatio.getKey().getName())
+                    .append(": \n* Coverage: ")
+                    .append(singleRatio.getValue())
+                    .append(" :white_check_mark:");
```
timja, line 54: consider putting this at the start of the line so that the emoji's are all aligned

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,104 @@
...
+    private String extractChecksText(CoverageResult result) {
+        Map<CoverageElement, Ratio> ratios = result.getResults();
+        Map<String, Float> lastRatios = getLastRatios(result);
+
+        StringBuilder text = new StringBuilder();
+        for (Map.Entry<CoverageElement, Ratio> singleRatio : ratios.entrySet()) {
+            text.append("## ")
+                    .append(singleRatio.getKey().getName())
+                    .append(": \n* Coverage: ")
+                    .append(singleRatio.getValue())
```
timja, line 53: coverage is normally summarised as a percentage, not out of the number of lines covered

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,104 @@
...
+    void publishChecks() {
+        ChecksPublisher publisher = ChecksPublisherFactory.fromRun(action.getOwner());
+        publisher.publish(extractChecksDetails());
+    }
+
+    private ChecksDetails extractChecksDetails() {
+        CoverageResult result = action.getResult();
+        ChecksOutputBuilder outputBuilder = new ChecksOutputBuilder()
+                .withTitle("Code Coverage")
+                .withSummary("## Generated " + result.getResults().size() + " types of coverage.")
```
timja, line 32: seems irrelevant to the user

pom.xml
```diff
@@ -36,7 +39,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.9</version>
+            <version>1.20</version>
```
timja, line 42: use the jenkins bom instead of choosing all these versions directly

https://github.com/jenkinsci/bom

you could introduce it in a separate PR to reduce the diff here as well

pom.xml
```diff
@@ -153,6 +171,10 @@
             <id>repo.jenkins-ci.org</id>
             <url>https://repo.jenkins-ci.org/public/</url>
         </repository>
+        <repository>
```
timja, line 174: shouldn't be needed,

run:
`mvn incrementals:incrementalify`

https://github.com/jenkinsci/incrementals-tools

pom.xml
```diff
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.9</version>
+        <version>4.0</version>
```
timja, line 7:
```suggestion
        <version>4.3</version>
```

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
XiongKezhi: `title` field in the `output` object. Then I'll change that part in warnings checks as well.

timja: take a look at this plugin: https://github.com/jenkinsci/github-coverage-reporter-plugin

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,129 @@
...
+ */
+class CoverageChecksPublisher {
+    private final CoverageAction action;
+
+    CoverageChecksPublisher(final CoverageAction action) {
+        this.action = action;
+    }
+
+    void publishChecks() {
+        ChecksPublisher publisher = ChecksPublisherFactory.fromRun(action.getOwner());
```
jeffpearce, line 24: Can this ever return null? If so, you don't want to do line 25

jeffpearce: Oh, I see. It returns a `NullChecksPublisher` if it can't find one

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,115 @@
...
+                        .append("%");
+            }
+
+            text.append("\n");
+        }
+
+        return text.toString();
+    }
+
+    private String extractChecksTitle(final CoverageResult result) {
```
uhafner, line 81: I would prefer to see line and branch coverage in the title.

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,115 @@
...
+                .withStatus(ChecksStatus.COMPLETED)
+                .withConclusion(ChecksConclusion.SUCCESS)
+                .withDetailsURL(action.getAbsoluteUrl())
+                .withOutput(output)
+                .build();
+    }
+
+    private String extractChecksText(final CoverageResult result) {
+        Map<CoverageElement, Ratio> ratios = result.getResults();
+        Map<CoverageElement, Ratio> lastRatios = getLastRatios(result);
```
uhafner, line 48: I wonder if it would make sense to separate the total report and the delta report? Maybe we can discuss that topic in our meeting today. This question is also relevant for the analysis report (new and total warnings).

src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
```diff
@@ -92,12 +92,18 @@ public void perform(@Nonnull Run<?, ?> run, @Nonnull FilePath workspace, @Nonnul
...
+        CoverageAction action = null;
         try {
-            processor.performCoverageReport(reportAdapters, reportDetectors, globalThresholds);
+            action = processor.performCoverageReport(reportAdapters, reportDetectors, globalThresholds);
         } catch (CoverageException e) {
             listener.getLogger().println(ExceptionUtils.getFullStackTrace(e));
             run.setResult(Result.FAILURE);
         }
+
+        if (action != null) {
```
uhafner, line 103: Move inside the `try` then `action = null` and the `if` is not required at all.

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -89,14 +89,15 @@ public CoverageProcessor(@Nonnull Run<?, ?> run, @Nonnull FilePath workspace, @N
...
      */
-    public void performCoverageReport(List<CoverageReportAdapter> reportAdapters, List<ReportDetector> reportDetectors, List<Threshold> globalThresholds)
+    public CoverageAction performCoverageReport(List<CoverageReportAdapter> reportAdapters, List<ReportDetector> reportDetectors, List<Threshold> globalThresholds)
             throws IOException, InterruptedException, CoverageException {
         Map<CoverageReportAdapter, List<CoverageResult>> results = convertToResults(reportAdapters, reportDetectors);
 
         CoverageResult coverageReport = aggregateReports(results);
         if (coverageReport == null) {
-            return;
+            return null;
```
uhafner, line 100: This looks somewhat strange (but this is from the existing code). When does the report 
- return `null`
- throw an `CoverageException`

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,115 @@
...
+
+        StringBuilder text = new StringBuilder();
+        for (Map.Entry<CoverageElement, Ratio> singleRatio : ratios.entrySet()) {
+            text.append("## ")
```
uhafner, line 52: Wouldn't it be simpler to have a table with two rows (titles and values) that contains a column for each type? Then it would not occupy so much place.

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,115 @@
...
+                .withOutput(output)
+                .build();
+    }
+
+    private String extractChecksText(final CoverageResult result) {
+        Map<CoverageElement, Ratio> ratios = result.getResults();
+        Map<CoverageElement, Ratio> lastRatios = getLastRatios(result);
+
+        StringBuilder text = new StringBuilder();
+        for (Map.Entry<CoverageElement, Ratio> singleRatio : ratios.entrySet()) {
```
uhafner, line 51: What is the default sorting? It looks ok in the PR results but I'm not sure if this is just some coincidence.

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,115 @@
...
+        ChecksOutput output = new ChecksOutputBuilder()
+                .withTitle(extractChecksTitle(result))
+                .withSummary("")
+                .withText(extractChecksText(result))
+                .build();
+
+        return new ChecksDetailsBuilder()
+                .withName("Code Coverage")
+                .withStatus(ChecksStatus.COMPLETED)
+                .withConclusion(ChecksConclusion.SUCCESS)
```
uhafner, line 40: The coverage report also has a quality gate that we should check.

src/test/java/io/jenkins/plugins/coverage/CoverageChecksPublisherTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+        ChecksDetails expectedDetails = new ChecksDetailsBuilder()
+                .withName("Code Coverage")
+                .withStatus(ChecksStatus.COMPLETED)
+                .withConclusion(ChecksConclusion.SUCCESS)
+                .withDetailsURL("http://127.0.0.1:8080/job/pipeline-coding-style/job/PR-3/49/coverage")
+                .withOutput(new ChecksOutputBuilder()
+                        .withTitle("Line coverage of 60%.")
+                        .withSummary("")
+                        .withText("## Conditional\n* :white_check_mark: Coverage: 40%\n"
```
uhafner, line 35: (*Conditional* actually is the wrong name, but this is something that needs to be changed in the coverage API and not in your PR)

pom.xml
XiongKezhi: but our plugin is currently not released, so we need this to download the checks dependencies

timja: That’s automatically added by the maven profile that incrementals adds behind the scenes

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
XiongKezhi: Done, see: https://github.com/XiongKezhi/codingstyle/pull/3/checks?check_run_id=907434646

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
XiongKezhi: I refactored my code, see field `COVERAGE_TYPES` in `CoverageChecksPublisher`, so I think the default sorting order is no longer a problem for us.

pom.xml
```diff
@@ -104,7 +107,34 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>jackson2-api</artifactId>
-            <version>2.8.11.3</version>
+            <version>2.11.1</version>
+        </dependency>
+        <dependency>
+            <groupId>io.jenkins.plugins</groupId>
+            <artifactId>checks-api</artifactId>
+            <version>0.1.0-rc4.fc43ef9447c7</version>
```
timja, line 115: release this?

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
XiongKezhi: the table format: https://github.com/XiongKezhi/codingstyle/pull/3/checks?check_run_id=907434646

src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
XiongKezhi: I just tried to keep the `try-catch` section as small as possible for easier exception locating, so isn't this a concern?

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
XiongKezhi: I refactored my code to get `CoverageAction` directly from `run` instead of changing this `CoverageProcessor`.

XiongKezhi:
Now I've used the table format for coverage and trend, added health report and failed message, example: https://github.com/XiongKezhi/codingstyle/pull/3/checks?check_run_id=909846128

The last thing to do is to enable configuration.

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        return lineTitle + " " + branchTitle;
+    }
+
+    private String extractChecksTitle(final String elementName, final String targetBuildName,
+                                      final float coverage, final float coverageDiff) {
+        StringBuilder title = new StringBuilder()
+                .append(elementName)
+                .append(" coverage: ")
```
timja, line 126: I would remove this line, it's obvious you're talking about coverage and github is truncating it in the UI
```suggestion
```

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,190 @@
...
+                + trends;
+    }
+
+    private String extractChecksTitle(final CoverageResult result) {
+        Map<CoverageElement, Ratio> lastRatios = getLastRatios(result);
+
+        String lineTitle;
+        float lineCoverage = result.getCoverage(CoverageElement.LINE).getPercentageFloat();
+        if (result.getLinkToBuildThatWasUsedForComparison() != null) {
+            lineTitle = extractChecksTitle("Line", "target branch build", lineCoverage,
```
timja, line 101: more trimming
```suggestion
            lineTitle = extractChecksTitle("Line", "target branch", lineCoverage,
```

src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
```diff
@@ -54,6 +54,8 @@
     private boolean calculateDiffForChangeRequests = false;
     private boolean failBuildIfCoverageDecreasedInChangeRequest = false;
 
+    private boolean publishChecks = false;
```
timja, line 57: Don’t we want to default to publishing? Change to skipPublishingChecks?

XiongKezhi: either is ok, I'll change it

pom.xml
XiongKezhi: I'm ready to release it now, but we may need to refactor some part since we have to publish checks for  `Queue.WaitingItem` instead of `Run` for `queued` state, is that ok? (I believe nothing changes in the consumer side after the that)

src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
```diff
@@ -197,12 +197,12 @@ public boolean getFailBuildIfCoverageDecreasedInChangeRequest() {
     }
 
     @DataBoundSetter
-    public void setPublishChecks(boolean publishChecks) {
-        this.publishChecks = publishChecks;
+    public void setSkipPublishingChecks(boolean skipPublishingChecks) {
```
uhafner, line 200: Would it then make sense to use the same name in the warnings plugin as well? This will help users to write their pipeline scripts.

XiongKezhi:
@jeffpearce @cizezsy We have finished most of the work, do you have time to review it? :)

timja:
cc @tempora-mutantur

pom.xml
timja: can you fix this please?

XiongKezhi: ok, done now.

src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,199 @@
+package io.jenkins.plugins.coverage;
+
+import edu.hm.hafner.util.VisibleForTesting;
+import hudson.model.Run;
+import io.jenkins.plugins.checks.api.*;
+import io.jenkins.plugins.checks.api.ChecksOutput.ChecksOutputBuilder;
+import io.jenkins.plugins.checks.api.ChecksDetails.ChecksDetailsBuilder;
+import io.jenkins.plugins.coverage.targets.*;
+import io.jenkins.plugins.util.JenkinsFacade;
+import jnr.ffi.Struct;
```
timja, line 10: what's this for?

XiongKezhi: oops, unused import...

PR #172, opened 2020-08-09
Merge `dev` branch to `master`

uhafner:
Almost every Jenkins plugin uses the Git default branch `master` for development. Additional feature and bugfix branches use meaningful names. Typically there is no need for a separate `dev` branch, the [git-flow](https://nvie.com/posts/a-successful-git-branching-model/) branching concept seems to be too complex for Jenkins plugin development. So I would suggest to merge the `dev` branch back to `master` so we do not create confusion. Currently there are already some commits in `master` that are not in `dev`. 

What do you think? @cizezsy @tempora-mutantur @timja


tempora-mutantur:
Changed the default github branch to "master".

Thank you

PR #179, opened 2020-10-30
Also set a fail message if marking stage as unstable so that checks API correctly reports non-success

### LINKED ISSUE #176: failBuildIfCoverageDecreasedInChangeRequest doesn't cause checks to fail
When the github checks are published for coverage with PR diff checks, the check always shows "success", even if the coverage drops. This appears to be due to the check status being determined by a failure message having been set in the action, and `setFailMessage()` not being called in the code path that raises the exception for the PR coverage reduction.

See:
https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java#L260
and
https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java#L53

mrginglymus:
Hopefully fixes #176

### DIFF
--- a/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
+++ b/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
@@ -482,6 +482,7 @@ public class CoverageProcessor {
                 action.setFailMessage(String.format("Build failed because following metrics did not meet stability target: %s.", unstableThresholds.toString()));
                 throw new CoverageException(action.getFailMessage());
             } else {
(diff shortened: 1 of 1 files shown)


uhafner (approved):

mrginglymus:
@uhafner is there anything else needed from me for this?

uhafner:
Ah, I forgot about this since this is a repository I am not actively monitoring...

PR #187, opened 2021-03-17
Missing Cobertura source files due to hardcoded URI's in Cobertura xml files.

trash-80:
Cobertura places a hardcoded URI within it's own xml files which causes find failures if the Jenkins' job is renamed or moved.  Current code attempts to find the files, but because each xml may have a different base URI, the current find doesn't necessarily create a correct path to the files.
Added code in branch does a regex search to find the file given a known subpath.  The added code attempts to find the missing piece of the URI.

I found this hardcoded URI has also caused issue for other projects that use Cobertura.  This patch is a bandaid to work around the issue until Cobertura itself works with relative paths.


uhafner:
There is a conflict. Can you also please remove unrelated changes? GitHub Actions, Whitespace changes, etc.

trash-80:
Simplified diff in branch to only related changes.

trash-80:
Github isn't allowing me to reopen this, so submitting a new simplified request.

Issue #162, opened 2020-05-26
Last update makes Cobertura created files unreadable


trash-80:
https://github.com/jenkinsci/code-coverage-api-plugin/blob/71a75c6a7589c208d1ab3a6768bd53cddeb68301/src/main/java/io/jenkins/plugins/coverage/adapter/converter/DocumentConverter.java#L26

When using 1.1.4, we found that while trying to read Cobertura created XML files, the coverage plugin would take 9 hours to complete, vs 21 seconds for the Cobertura plugin itself.

Upon investigation, I found the reason was some of the settings introduced in 1.1.5 were necessary as the coverage plugin was timing out while trying to read 100's to 1000's of XML files.  It seemed to attempt to read twice, then fail.  Reading 3 files took over 5 minutes to fail for each.

When I updated to 1.1.7-Snapshot, the same area would fail sooner because of the above references code causes a file read to fail immediately.  The problem is: Covertura's most recent release (2.1.1) still adds DOCTYPE to it's files.

For testing, I changed the "disallow-doctype-decl" to false and our code builds in seconds again.

[SECURITY-1699] fix broke the Coverage Plugin for Cobertura created files.  Though, I can't find a comment in the commit as to why this one was necessary.

trash-80:
Here is the line in Cobertura's source:
```
println("<!DOCTYPE coverage SYSTEM \"http://cobertura.sourceforge.net/xml/"
					+ coverageDTD + "\">")
```
[https://github.com/cobertura/cobertura/blob/9de3fb15edd73141c8b6887f07840ec3250a62d1/cobertura/src/main/java/net/sourceforge/cobertura/reporting/xml/XMLReport.java#L66](url)

trash-80:
Cobertura current doesn't seem to have a way to turn off adding the DOCTYPE line, so by adding the above security feature, Cobertura reports won't parse.

Even if Cobertura corrected the DOCTYPE line, one of the goals of the Code Coverage plugin is to compare coverage across builds, thus previous builds would most likely have reports with the DOCTYPE added.

Per OWASP, this line isn't required if 3 other lines are added, while the 3 other lines are not as good defense as "disallow-doctype-decl", they understand not everyone can changeover immediatly.

The code I referenced in the OP is one of four, in which all 4 suggestions from OWASP were implemented, which means whomever suggested the edit didn't really read the page because not all four options are required at the same time.  Basically, the code as implemented is like adding a padlock to a safe's door.

[...]

PR #189, opened 2021-03-18
Ignore Doctype errors caused by Cobertura adding to top of xml files.

trash-80:
Fix for Issue: https://github.com/jenkinsci/code-coverage-api-plugin/issues/162

In our environment, we noticed that Cobertura files were failing consumption.  The error was traced to Cobertura itself adding a doctype to the top of it's files.  
From commit 0a43672:
"Per OWASP (https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#jaxp-documentbuilderfactory-saxparserfactory-and-dom4j), not all four features need set at the same time.
If "disallow-doctype-decl" cannot be set to false, in this case because Cobertura sets the DOCTYPE, then at least "external-general-entities" should be set to false.
Both do not need to be set as "disallow-doctype-decl" is already a more secure setting, making setting "external-general-entities" worthless."


uhafner (approved):
👍  I also deactivated this option in my warnings plugin.

Issue #182, opened 2021-02-01
Use echarts-api plugin


timja:
Remove our local copy of echarts:

https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/src/main/webapp/scripts/echarts.min.js

and switch to using the [API plugin](https://github.com/jenkinsci/echarts-api-plugin)

That will mean only one copy of echarts per page, and means we don't have to maintain echarts etc

fyi @uhafner

uhafner:
Yes, this is already tracked as https://issues.jenkins.io/browse/JENKINS-62633. 

Two issue trackers are not really useful...

timja:
Only 5 open issues there, and none raised by users, probably should just close them all and maybe delete the component

simonsymhoven:
closes #197.

PR #195, opened 2021-05-20
Add coverage portlet for pull-request-monitoring plugin

simonsymhoven:
Implement portlet for the https://github.com/jenkinsci/pull-request-monitoring-plugin. The portlet visualize the aggregated results for line, method and instruction coverage of the CoverageResult for the CoverageAction. I used the existing diagram of the CoverageResult as a guideline and implemented the following portlet based on the existing API for the different cases:

Bildschirmfoto 2021-05-26 um 13 34 25


simonsymhoven:
Due to #182 and failing checks because of Jenkins baseline version (see #196), marked this pr as draft again. Will come back later after finished #182 and extended the change-request coverage api to implement better portlet for pull-request-monitoring plugin.

src/main/java/io/jenkins/plugins/coverage/CoveragePullRequestMonitoringPortlet.java
```diff
@@ -0,0 +1,122 @@
...
+     * @return
+     *          the transformed list.
+     */
+    @JavaScriptMethod
+    public List<CoverageResult.JSCoverageResult> getResults() {
+        List<CoverageResult.JSCoverageResult> results = new LinkedList<>();
+
+        for (Map.Entry<CoverageElement, Ratio> c : action.getResult().getResults().entrySet()) {
+            String name = c.getKey().getName();
+            if ("Conditional".equals(name) || "Line".equals(name) || "Instruction".equals(name)) {
```
uhafner, line 77: It seems a little bit fragile to compare by string. I think `LINE` and `CONDITIONAL` are constants and can be reused. 

And I rarely use instruction since normally one is looking at the source code line by line. So for me it would be sufficient to use only line and branch (which is called conditional here).

src/main/resources/io/jenkins/plugins/coverage/CoveragePullRequestMonitoringPortlet/monitor.jelly
```diff
@@ -0,0 +1,48 @@
...
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler">
+
+    <st:adjunct includes="io.jenkins.plugins.echarts"/>
+    <st:adjunct includes="io.jenkins.plugins.bootstrap5"/>
+
+    <j:set var="targetBranchBuild" value="${it.getComparedBuildLink()}" />
+    <j:if test="${targetBranchBuild.isPresent()}">
+        <j:set var="coverageDiff" value="${it.getDiff()}" />
+        <j:choose>
+            <j:when test="${coverageDiff gt 0}">
```
uhafner, line 12: I wonder if we can use the code in `Ratio` to render the percentage.

(In the screenshots it looks like that sometimes for a value of 0.000001 this condition is true, maybe we should compare after rounding)

simonsymhoven:
I started to wrap the interfaces into abstract classes. Will come back here and update the pr after https://github.com/jenkinsci/pull-request-monitoring-plugin/pull/75 finished and released.

uhafner (approved):

PR #198, opened 2021-05-22
Extend delta coverage api

simonsymhoven:
For now only the line coverage is calculating as diff, if `calculateDiffForChangeRequests` is enabled and a target branch is found.
* Extended  the diff coverage api by all existing coverage elements of target branch. 
* make variables names more meaningful: 
1. `changeRequestCoverageDiffWithTargetBranch` does not exist anymore and is part of the deltaResult map now. The getter, which is used until now `getChangeRequestCoverageDiffWithTargetBranch` is replaced by `getLineCoverageDelta()`, which extracts the corresponding metric from the deltaResults map. Therefore other plugins are able to use all delta metrics.
2. `linkToBuildThatWasUsedForComparison`is now the `referenceBuildUrl`

I'm not sure if this is all working as expected or if I'm breaking the api with it.. So can this please be checked carefully. Thanks in advance. Unit tests are still passing.

### DIFF
--- a/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
+++ b/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
@@ -210,9 +210,9 @@ public class CoverageProcessor {
     }
 
     private void failBuildIfChangeRequestDecreasedCoverage(CoverageResult coverageResult) throws CoverageException {
-        float coverageDiff = coverageResult.getLineCoverageDelta();
+        float coverageDiff = coverageResult.getCoverageDelta(CoverageElement.LINE);
         if (coverageDiff < 0) {
-            throw new CoverageException("Fail build because this change request decreases code coverage by " + coverageDiff);
+            throw new CoverageException("Fail build because this change request decreases line coverage by " + coverageDiff);
         }
     }
 
--- a/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
+++ b/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
@@ -40,6 +40,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.TreeMap;
 import java.util.stream.Collectors;
 
 public class CoverageProcessor {
@@ -193,20 +194,23 @@ public class CoverageProcessor {
             return;
         }
 
-        // Calculate diff only for line coverage
-        Ratio changeRequestLinesCoverage = coverageReport.getCoverage(CoverageElement.LINE);
-        if (changeRequestLinesCoverage == null) {
-            return;
-        }
+        Map<CoverageElement, Float> deltaCoverage = new TreeMap<>();
+        targetBranchCoverageResult.getResults().forEach((coverageElement, referenceRatio) -> {
+            Ratio buildRatio = coverageReport.getCoverage(coverageElement);
 
-        float percentageDiff =
-                changeRequestLinesCoverage.getPercentageFloat() - referenceLineCoverage.getPercentageFloat();
-        coverageReport.setChangeRequestCoverageDiffWithTargetBranch(percentageDiff);
-        coverageReport.setLinkToBuildThatWasUsedForComparison(referenceBuild.getUrl());
+            if (buildRatio != null) {
+                float diff = buildRatio.getPercentageFloat() - referenceRatio.getPercentageFloat();
+                listener.getLogger().println(coverageElement.getName() + " coverage diff: " + diff + "%. Add to CoverageResult.");
+                deltaCoverage.put(coverageElement, diff);
+            }
+        });
+
+        coverageReport.setReferenceBuildUrl(buildToTakeCoverageFrom.getUrl());
+        coverageReport.setDeltaResults(deltaCoverage);
     }
 
     private void failBuildIfChangeRequestDecreasedCoverage(CoverageResult coverageResult) throws CoverageException {
-        float coverageDiff = coverageResult.getChangeRequestCoverageDiffWithTargetBranch();
+        float coverageDiff = coverageResult.getLineCoverageDelta();
         if (coverageDiff < 0) {
             throw new CoverageException("Fail build because this change request decreases code coverage by " + coverageDiff);
         }
--- a/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
+++ b/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
@@ -195,7 +195,7 @@ public class CoverageProcessor {
         }
 
         Map<CoverageElement, Float> deltaCoverage = new TreeMap<>();
-        targetBranchCoverageResult.getResults().forEach((coverageElement, referenceRatio) -> {
+        referenceCoverageResult.getResults().forEach((coverageElement, referenceRatio) -> {
             Ratio buildRatio = coverageReport.getCoverage(coverageElement);
 
             if (buildRatio != null) {
@@ -205,7 +205,7 @@ public class CoverageProcessor {
             }
         });
 
-        coverageReport.setReferenceBuildUrl(buildToTakeCoverageFrom.getUrl());
+        coverageReport.setReferenceBuildUrl(referenceBuild.getUrl());
         coverageReport.setDeltaResults(deltaCoverage);
     }
 
--- a/src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
+++ b/src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
@@ -77,6 +77,7 @@ public class CoverageResult implements Serializable, Chartable, ModelObject {
     private String tag;
 
     private String referenceBuildUrl = null;
+    private float changeRequestCoverageDiffWithTargetBranch = 0;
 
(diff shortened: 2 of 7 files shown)


src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
```diff
@@ -77,11 +76,7 @@
     private String name;
     private String tag;
 
-    /**
-     * Properties to store a change request coverage diff with the latest successful build from the target branch
-     */
-    private float changeRequestCoverageDiffWithTargetBranch = 0;
-    private String linkToBuildThatWasUsedForComparison = null;
+    private String referenceBuildUrl = null;
```
uhafner, line 79: Renaming fields (and public getters) normally is not a good thing:
- it breaks serialization (if no migration is provided)
- it breaks callers (or subclasses)

However, since the serialization uses Java serialization (https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java#L691) and not the XStream serialization previous results will be broken anyway if we add new content. So it might be ok here. 

@oleg-nenashev @cizezsy what was the rationale behind that decision not to use Jenkins default XML serialization in this plugin?

src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
```diff
@@ -92,6 +87,8 @@
 
     private final Map<CoverageElement, Ratio> localResults = new TreeMap<>();
 
+    private Map<CoverageElement, Float> deltaResults = new TreeMap<>();
```
uhafner, line 90: Shouldn't the values be of type `Ratio` again? (Maybe we need to add a `subtract` method in `Ratio`)

src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
```diff
@@ -231,39 +228,31 @@ private String sanitizeFilename(String inputName) {
...
     /**
-     * Getter for property 'changeRequestCoverageDiffWithTargetBranch'.
+     * Getter for property 'lineCoverageDelta'.
      *
-     * @return Value for property 'changeRequestCoverageDiffWithTargetBranch'.
+     * @return Value for property 'lineCoverageDelta'.
      */
-    public float getChangeRequestCoverageDiffWithTargetBranch() {
-        return changeRequestCoverageDiffWithTargetBranch;
+    public float getLineCoverageDelta() {
```
uhafner, line 235: It would be saver to mark the old method as `@Deprecated` and add the new one as an alias.

src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
```diff
@@ -374,6 +363,24 @@ public CoverageElement getChildElement() {
...
+        return Collections.unmodifiableMap(deltaResults);
+    }
+
+    /**
+     * Setter for property 'deltaResults'.
+     *
+     * @param deltaResults Value to set for property 'deltaResults'.
+     */
+    public void setDeltaResults(Map<CoverageElement, Float> deltaResults) {
+        this.deltaResults = deltaResults;
```
uhafner, line 381: Since the map is already there it would be saver to change the values of the map rather then changing the map.

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -236,29 +229,36 @@ private void setDiffInCoverageForChangeRequest(CoverageResult coverageReport) {
...
-                changeRequestLinesCoverage.getPercentageFloat() - targetBranchLinesCoverage.getPercentageFloat();
-        coverageReport.setChangeRequestCoverageDiffWithTargetBranch(percentageDiff);
-        coverageReport.setLinkToBuildThatWasUsedForComparison(buildToTakeCoverageFrom.getUrl());
+
+        Map<CoverageElement, Float> deltaCoverage = new TreeMap<>();
+        targetBranchCoverageResult.getResults().forEach((coverageElement, referenceRatio) -> {
+            Ratio buildRatio = coverageReport.getCoverage(coverageElement);
+
+            if (buildRatio != null) {
+                float diff = buildRatio.getPercentageFloat() - referenceRatio.getPercentageFloat();
```
uhafner, line 250: See comment below.

uhafner (changes requested):

src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
simonsymhoven: I'm not sure, if I get this point.. What exactly do you mean with this?

uhafner: Effective Java Item 50: Always make defensive copies, never store mutable objects in fields. 

```suggestion
        this.deltaResults.clear();
        this.deltaResults.addAll(deltaResults);
```

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
simonsymhoven: Which comment below?

src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
simonsymhoven: I am not sure if this is needed. The line coverage diff was saved as float until now as well. Also I am not sure how to calculate the diff between two Ratios with numerator and denominator. What exactly do they express? The diff between the two builds was calculated with the `getPercentageFloat()` method, so the coverage is calculated as `(100 * numerator / denominator)`. So if we want to save the diff as Ratio too, we basically need to subtract the numerator of Ratio1 from Ratio2 and the denominator from Ratio1 from Ratio2 and divide it, or am I forgetting something?

uhafner: I just thought since we already have that class that handles rounding, formatting, etc. why not reuse it for the delta. And to subtract two rational numbers we need both numbers use the common denominator and subtract the numerators afterwards. But if we do not need the details of `Ratio` we simply can use float values, that is fine. (When creating a delta we would also need to make `Ratio` sign aware which it is currently not).

src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
uhafner: Ah, actually above, that `Ratio` vs. `float` discussion.

src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
simonsymhoven: That might be the nicer solution. I will check it out and try to use the Ratio.

simonsymhoven: It doesn't really make sense to me... Have now left it like this and implemented it with floats.

src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
```diff
@@ -230,10 +228,23 @@ private String sanitizeFilename(String inputName) {
...
+        return deltaResults.getOrDefault(element, 0.0F);
+    }
+
     /**
      * Getter for property 'changeRequestCoverageDiffWithTargetBranch'.
      *
      * @return Value for property 'changeRequestCoverageDiffWithTargetBranch'.
+     * @deprecated use {@link #getCoverageDelta(CoverageElement)} instead.
      */
     public float getChangeRequestCoverageDiffWithTargetBranch() {
```
uhafner, line 249:
```suggestion
     * @deprecated use {@link #getCoverageDelta(CoverageElement)} instead.
     */
    @Deprecated
    public float getChangeRequestCoverageDiffWithTargetBranch() {
```

src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
```diff
@@ -243,27 +254,28 @@ public float getChangeRequestCoverageDiffWithTargetBranch() {
      * Setter for property 'changeRequestCoverageDiffWithTargetBranch'.
      *
      * @param changeRequestCoverageDiffWithTargetBranch Value to set for property 'changeRequestCoverageDiffWithTargetBranch'.
+     * @deprecated diff coverage is stored in {@link #deltaResults}.
      */
     public void setChangeRequestCoverageDiffWithTargetBranch(float changeRequestCoverageDiffWithTargetBranch) {
```
uhafner, line 259:
```suggestion
     * @deprecated diff coverage is stored in {@link #deltaResults}.
     */
    @Deprecated
    public void setChangeRequestCoverageDiffWithTargetBranch(float changeRequestCoverageDiffWithTargetBranch) {
```

uhafner (changes requested):
It helps to use `@Drepecated` as well.

src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java
simonsymhoven: Oh for sure, i missed that..

uhafner (approved):

PR #203, opened 2021-08-12
User interface refresh and code cleanup

uhafner:
# User Interface Refresh 🌟 

This PR will refresh the user interface of the coverage plugin. It will reuse several elements of the warnings plugin so that both plugins share a similar visualization. Additionally, this PR creates a new model that is decoupled from the user interface. The results are now also stored using Jenkins default XML serialization, otherwise keeping the results backwards compatible during the plugin evolution will not be possible. Since the existing code base is rather big, not all parts of the plugin have been refactored so far. In order to see the current progress, I decided to keep the old data structures in parallel. So if the new things break something, teams can switch back to the old version without affecting the results. Technically this has been implemented by creating the old *and* new action results for every build. That means, that the new actions will show up for new builds and the old actions will be created but hidden in the UI. For old builds, the old actions still show up. In case you need to revert to the old version of the plugin, then the hidden old results will reappear. If this PR is stable enough we can change that behavior.

The following UI elements have been added and changed. These building blocks are now almost ready from a feature perspective. Now it would make sense to test these elements and improve them based on your user experience. 

## Trend chart

Replace the build information chart (coverage statistics bar chart) with a job trend chart that shows the line and branch coverage for the selected number of builds. It reuses the new chart configuration dialog of the Bootstrap 5 and ECharts plugins.

Trend Chart

Open Tasks:
- [ ] Create a trend chart for delta values
- [ ] Reuse trend chart for details view

## Delta Computation

Delta computation is done using the configured reference build (see [forensics-api-plugin](https://github.com/jenkinsci/forensics-api-plugin)). If no reference build is found, then the previous build of the same job will be considered. 

## Build summary

Add actions on the job *and* build level to navigate to the latest coverage details. Replace the tango icon with a modern FontAwesome icon. Focus on the interesting values in the build summary card. Additional information is available in the details view.

Build Summary

## Build details

Bootstrap cards are now used to show the elements in the detail views to provide a responsive design. 
The following elements are available. 

### Summary

In the summary one now sees immediately the results of the project. Empty hierarchy nodes are removed.  

Summary

### Details per file

A new tree map graphically shows the coverage per file. You can zoom into packages and see the coverage percentage of individual files.

Details Treemap

### Files table 

A table with all projects files shows the details for the source files. From there one can jump into the source file view. This table is inspired by the warnings plugin. 

Table

### Source table 

In the source view (if source code scanning has been enabled) the detailed coverage of the source code is shown.

Source View

## Next steps

- [x] Create a helpful layout for the elements in the build details view (tabs, small or large screen layouts, etc.)
- [ ] Add a view for multiple projects (or reuse the same view and use the project as module like in the warnings plugin?)

## Chore

- [x] Cleanup code and package structure
- [x] Write more tests


oleg-nenashev:
Looks great! Let me know when ready for testing

uhafner:
> Looks great! Let me know when ready for testing

Yes, this will take some more iteration steps... I'm on the road the next three weeks so I'm not sure how much time I can spend. Since I am also changing the serialization I want to make sure that everything works so we won't need to change it again (and loose the job histories multiple times).

In October a student will start his master thesis and will try improve the delta coverage handling (pull request, changed files vs. all files, etc.). So until then I will have the new views in good shape so the student will have a good starting point.

uhafner:
So I think I implemented now the building blocks of this PR. I still need to cleanup the code base and write some more test. But I think it is already feature complete so others can test it. 

I also made a demo at the last UI SIG meeting, the recording is available here: https://community.jenkins.io/t/ui-overhaul-of-code-coverage-plugin/363/5.

uhafner:
I finished most of my planned  rework now. Since nobody commented so far (actually the community is really quiet in the moment) my plan is to merge this PR at the end of the week. Then I will create a new release so others have a chance to see the changes. Especially for the Saturday Hacktoberfest event I want to have the new plugin version available, so potential contributors can see the new UI in action. As noted in the PR description, the results are stored in the new **and** old format, so a fallback to the previous release should be possible if something breaks. 

@oleg-nenashev @timja If you have objections to this plan please let me know. Then I will only merge but do not release the changes.

timja:
no objections, I've been following it :)

balderbach:
> I finished most of my planned rework now. Since nobody commented so far (actually the community is really quiet in the moment) my plan is to merge this PR at the end of the week.

This looks super-promising, thanks for all this work and for the great screenshot updates shared to go along with it. Will recommend our organization to try this out, and the screenshots are especially helpful to feel motivated to start :+1:

PR #232, opened 2021-10-14
Do not fail when report is missing in reference build

zbynek:
- [n/a] Link to relevant issues in GitHub or Jira
- [n/a] Link to relevant pull requests, esp. upstream and downstream changes
- [TODO] Ensure you have provided tests - that demonstrates feature works or fixes the issue

As an administrator I want to be able to remove the coverage .xml files for old builds without breaking new builds. Clearly this should be done using some log-rotate feature in Jenkins, but it would be nice if the plugin could handle the case when I do it by simply removing the files from the filesystem.


zbynek:
@uhafner I didn't find any test coverage for the diff handling. Do you expect a test for this PR?

PR #245, opened 2021-11-04
Remove agent-to-controller calls

jglick:
Compatible with https://github.com/jenkinsci/jenkins/pull/5885. See https://github.com/jenkinsci/remoting-security-workaround-plugin/blob/main/README.adoc#affected-functionality.

### DIFF
--- a/src/test/java/io/jenkins/plugins/coverage/BlockSlaveToMasterFileCallable.java
+++ b/src/test/java/io/jenkins/plugins/coverage/BlockSlaveToMasterFileCallable.java
@@ -1,0 +1,52 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2021 CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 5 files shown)


timja (approved):

src/test/java/io/jenkins/plugins/coverage/BlockSlaveToMasterFileCallable.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Prevents all {@link SlaveToMasterFileCallable}s from running during tests, to make sure we do not rely on them.
+ */
+public class BlockSlaveToMasterFileCallable extends ReflectiveFilePathFilter {
+
+    @Override protected boolean op(String name, File path) throws SecurityException {
+        throw new SecurityException("refusing to " + name + " on " + path);
+    }
+
+    @Extension public static class ChannelConfiguratorImpl extends ChannelConfigurator {
```
uhafner, line 44: Shouldn't this be a `TestExtension`?

uhafner (approved):

src/test/java/io/jenkins/plugins/coverage/BlockSlaveToMasterFileCallable.java
jglick: No, because that would only apply within a single test. This deliberately gets loaded in all tests in the plugin.

PR #265, opened 2021-11-28
Rewrite source code painting and archiving 

uhafner:
The current approach of painting and archiving source files is very slow and has several flaws.

This new algorithm rewrites this part of the code from scratch.

Here is a sketch of the new algorithm:
- Painting is complete done in a single step on the controller
- Relative paths are automatically determined from the coverage tree structure
- Configurable source path prefix let the callers decide where the source are located
- Painted files are written into a folder in one step (on the agent)
- Painted files folder is zipped and copied in one call from the agent to the controller

Open tasks:
- [ ] Make encoding of source files a parameter of the publisher
- [ ] Make the source path folder(s) a parameter of the publisher
- [ ] Combine scanning for report files, parsings into model and painting of sources in one single step without additional data transfer between controller and agent
- [ ] Use new files in source view
- [ ] Check if we can use prism in source code syntax highlighting 
- [ ] ZIP individual sources as well and decompress on access only
- [ ] Cleanup
- [ ] Tests

### DIFF
--- a/src/test/java/io/jenkins/plugins/coverage/model/JavaGitContainer.java
+++ b/src/test/java/io/jenkins/plugins/coverage/model/JavaGitContainer.java
@@ -1,0 +1,11 @@
+package io.jenkins.plugins.coverage.model;
+
+import org.jenkinsci.test.acceptance.docker.DockerFixture;
+import org.jenkinsci.test.acceptance.docker.fixtures.SshdContainer;
+
+@DockerFixture(id = "java-git", ports = {22, 8080}
+)
+public class JavaGitContainer extends SshdContainer {
+    public JavaGitContainer() {
+    }
+}
--- a/src/test/java/io/jenkins/plugins/coverage/model/JavaGitContainer.java
+++ b/src/test/java/io/jenkins/plugins/coverage/model/JavaGitContainer.java
@@ -3,9 +3,18 @@ package io.jenkins.plugins.coverage.model;
 import org.jenkinsci.test.acceptance.docker.DockerFixture;
 import org.jenkinsci.test.acceptance.docker.fixtures.SshdContainer;
 
-@DockerFixture(id = "java-git", ports = {22, 8080}
-)
+/**
+ * Represents a server with SSHD, Java tooling, Maven and Git.
+ *
+ * @author Ullrich Hafner
+ */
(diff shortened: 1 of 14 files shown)


src/main/java/io/jenkins/plugins/coverage/CoveragePublisher.java
```diff
@@ -1,49 +1,60 @@
...
+import io.jenkins.plugins.prism.SourceCodeDirectory;
+import io.jenkins.plugins.util.JenkinsFacade;
 
 public class CoveragePublisher extends Recorder implements SimpleBuildStep {
-
     private List<CoverageAdapter> adapters = new LinkedList<>();
     private List<Threshold> globalThresholds = new LinkedList<>();
 
+    private String sourceCodeEncoding = StringUtils.EMPTY;
+    private Set<SourceCodeDirectory> sourceDirectories = new HashSet<>(); // @since 9.11.0
```
uhafner, line 56: Wrong version

src/main/java/io/jenkins/plugins/coverage/model/CoverageBuildAction.java
```diff
@@ -178,30 +182,18 @@ protected String getBuildResultBaseName() {
...
 
     @Override
     public CoverageViewModel getTarget() {
         return new CoverageViewModel(getOwner(), getResult());
     }
 
-    public HealthReport getHealthReport() {
-        return healthReport;
-    }
-
```
uhafner, line 192: Remove setter and make final

src/test/java/io/jenkins/plugins/coverage/model/JavaGitContainer.java
```diff
@@ -0,0 +1,22 @@
...
+import org.jenkinsci.test.acceptance.docker.DockerFixture;
+import org.jenkinsci.test.acceptance.docker.fixtures.SshdContainer;
+
+/**
+ * Represents a server with SSHD, Java tooling, Maven and Git.
+ *
+ * @author Ullrich Hafner
+ */
+@DockerFixture(id = "java-git", ports = {22, 8080})
+public class JavaGitContainer extends SshdContainer {
```
uhafner, line 12: Move to integration test of plug-in util

PR #267, opened 2021-12-02
[Issue 67148] - Create integration tests for freestyle job and pipelines

bleibdirtroy:
Tests which are annotated with `@Ignore`:

`CodeCoveragePluginITest`:
- `freestyleForNoAdapter()` -> build should fail if no adapter is provided
- `pipelineForNoAdapter()` -> build should fail if no adapter is provided
- `freestyleForTwoCobertura()` -> returns different line-coverage depending on using cobertura files within a pipeline or freestyle project (no difference with jacoco files)
 
`DeltaComputationVsReferenceBuildITest`:
- `freestyleProjectTryCreatingReferenceBuildWithDeltaComputation()` -> delta is different depending on using freestyle or pipeline job.

`MultipleInvocationsOfStepITest`:
- `withDifferentTag()` -> should return two CoverageBuildActions but returns one.

`ReportAggregationITest`:
- `freeStyleProjectCheckReportAggregationWithCoberturaFiles()` -> returns different line-coverage depending on using cobertura files within a pipeline or freestyle project (no difference with jacoco files)

### DIFF
--- a/src/test/java/io/jenkins/plugins/coverage/model/CoveragePluginITest.java
+++ b/src/test/java/io/jenkins/plugins/coverage/model/CoveragePluginITest.java
@@ -30,7 +30,7 @@ public class CoveragePluginITest extends IntegrationTestWithJenkinsPerSuite {
     private static final String JACOCO_FILE_NAME_2 = "jacoco-codingstyle.xml";
 
     private static final String COBERTURA_FILE_NAME = "cobertura-higher-coverage.xml";
-    private static final String COBERTURA_FILE_NAME_2 = "coverage-with-lots-of-data.xml";
+    private static final String COBERTURA_FILE_NAME_2 = "../coverage-with-lots-of-data.xml";
 
     /** Example integration test for a pipeline with code coverage. */
     @Test
--- a/src/test/java/io/jenkins/plugins/coverage/model/CoveragePluginITest.java
+++ b/src/test/java/io/jenkins/plugins/coverage/model/CoveragePluginITest.java
@@ -26,9 +26,16 @@ import static org.assertj.core.api.Assertions.*;
  */
 public class CoveragePluginITest extends IntegrationTestWithJenkinsPerSuite {
 
+    public static final int JACOCO_MISSED_LINES = 6083;
+    public static final int JACOCO_ALL_LINES = 6368;
+    public static final int BOTH_JACOCO_MISSED_LINES = 6377;
+    public static final int BOTH_JACOCO_ALL_LINES = 6691;
+    public static final int COBERTURA_COVERED_LINES = 2;
+    public static final int COBERTURA_ALL_LINES = 2;
+    public static final int JACOCO_COBERTURA_COVERED_LINES = 6085;
+    public static final int JACOCO_COBERTURA_ALL_LINES = 6370;
     private static final String JACOCO_ANALYSIS_MODEL_FILE = "jacoco-analysis-model.xml";
     private static final String JACOCO_CODINGSTYLE_FILE = "jacoco-codingstyle.xml";
-
     private static final String COBERTURA_HIGHER_COVERAGE_FILE = "cobertura-higher-coverage.xml";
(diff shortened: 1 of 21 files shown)


src/test/java/io/jenkins/plugins/coverage/model/CoveragePluginITest.java
```diff
@@ -0,0 +1,335 @@
...
+/**
+ * Integration test for different jacoco and cobertura files.
+ */
+public class CoveragePluginITest extends IntegrationTestWithJenkinsPerSuite {
+
+    private static final String JACOCO_FILE_NAME = "jacoco-analysis-model.xml";
+    private static final String JACOCO_FILE_NAME_2 = "jacoco-codingstyle.xml";
+
+    private static final String COBERTURA_FILE_NAME = "cobertura-higher-coverage.xml";
+    private static final String COBERTURA_FILE_NAME_2 = "coverage-with-lots-of-data.xml";
```
uhafner, line 33: This file is one level higher: either copy it to the model package or use path `../coverage-with-lots-of-data.xml`

src/test/java/io/jenkins/plugins/coverage/model/CoveragePluginSkipPublishingOfChecksITest.java
```diff
@@ -0,0 +1,183 @@
...
+    @Test
+    public void pipelinePublishingOfChecksWithRepo() {
+        checkConsoleLog(getPipelineProjectWithSCM(false), false);
+    }
+
+    /**
+     * Tests publishing of checks with source code when skip publishing checks is true.
+     */
+    @Test
+    public void freeStyleSkipPublishingOfChecksWithRepo() throws IOException {
```
uhafner, line 87: This test will erase your coverage file and will skip the checks because there is no result

src/test/java/io/jenkins/plugins/coverage/model/CoveragePluginSkipPublishingOfChecksITest.java
```diff
@@ -0,0 +1,183 @@
...
+     *
+     * @param build
+     *         the successful build
+     * @param skipPublishingChecks
+     *         if publishing checks should be skipped
+     */
+    private void checkConsoleLog(final Run<?, ?> build, final boolean skipPublishingChecks) {
+        String consoleLog = getConsoleLog(build);
+        if (skipPublishingChecks) {
+            assertThat(consoleLog).contains("Skip publishing Coverage report....");
```
uhafner, line 113: You need to check for absence of
```
[Checks API] No suitable checks publisher found.
```

src/test/java/io/jenkins/plugins/coverage/model/CoveragePluginSkipPublishingOfChecksITest.java
```diff
@@ -0,0 +1,183 @@
...
+        }
+        else {
+            assertThat(consoleLog).contains("Publishing Coverage report....");
```
uhafner, line 116: and her for presence of the string
```
[Checks API] No suitable checks publisher found.
```

src/test/java/io/jenkins/plugins/coverage/model/CoverageSourceCodeRenderingAndCopyingITest.java
```diff
@@ -0,0 +1,136 @@
...
+    private void verifyGitRepository(final WorkflowJob workflowJob) {
+        Run<?, ?> build = buildSuccessfully(workflowJob);
+
+        String consoleLog = getConsoleLog(build);
+
+        assertThat(consoleLog)
+                .contains("Cloning repository " + REPOSITORY)
+                .contains("Checking out Revision " + COMMIT)
+                .contains("git checkout -f " + COMMIT);
+    }
```
uhafner, line 89: This verifies the Git cloning but what we need is verifying the source code copying and visualization

uhafner (changes requested):

PR #286, opened 2022-01-10
This is a change from Float to BigDecimal to get more precision in th…

pgermosen:
…73.4 % of coverage

### DIFF
--- a/src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
+++ b/src/main/java/io/jenkins/plugins/coverage/CoverageChecksPublisher.java
@@ -1,5 +1,6 @@
 package io.jenkins.plugins.coverage;
 
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -116,7 +117,7 @@ class CoverageChecksPublisher {
             float lineCoverage = result.getCoverage(CoverageElement.LINE).getPercentageFloat();
             if (result.getReferenceBuildUrl() != null) {
                 title.append(extractChecksTitle("Line", "target branch", lineCoverage,
-                        result.getCoverageDelta(CoverageElement.LINE)));
+                        result.getCoverageDelta(CoverageElement.LINE).floatValue()));
             } else if (lastRatios.containsKey(CoverageElement.LINE)) {
                 title.append(extractChecksTitle("Line", "last successful build", lineCoverage,
                         lineCoverage - lastRatios.get(CoverageElement.LINE).getPercentageFloat()));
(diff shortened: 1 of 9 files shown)


uhafner:
Hmm, did you notice actual problems with the current approach? The coverage typically is measured in integers (missed, coverage elements) anyway so I do not see much benefit into changing them into values that just might have a better precision. Only the actual ratio is shown in the UI as float (covered/total). Changing the type of API elements is quite a complex step because this will break the serialization.

pgermosen:
> nyway so I do not see much benefit by changing them into values that just might have a better precision. Only the actual ratio is shown in the UI as float (covered/total). Changing the type of API elements is quite a complex step because this will break the serialization.

Ok, So is not a great solution and it breaks serialization (that means... I could not be implemented as it is).
But I still want to have fewer false positives, of pull requests with no changes, and look at a lower coverage because of loss of precision.

Using a float type is not a good approach, for this kind of information. what can we do to make it better?

uhafner:
I think we can change the way that the delta is computed but not the way the coverage is stored. As noted above the coverage is using integers in the new model: https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/src/main/java/io/jenkins/plugins/coverage/model/Coverage.java

So we actually need to change https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java#L250..L259 only. (I'm not sure if that is possible without changing https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java as well. The problem is that changing `CoverageResult` may break the serialization of old results.) 

Currently we are replacing the old model `CoverageResult` with the new model `CoverageNode` step by step. I think the model of the new delta computation (that is work in progress) will handle the problem you noticed automatically.

pgermosen:
> I think we can change the way that the delta is computed but not the way the coverage is stored. As noted above the coverage is using integers in the new model: https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/src/main/java/io/jenkins/plugins/coverage/model/Coverage.java
> 
> So we actually need to change https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java#L250..L259 only. (I'm not sure if that is possible without changing https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/src/main/java/io/jenkins/plugins/coverage/targets/CoverageResult.java as well. The problem is that changing `CoverageResult` may break the serialization of old results.)
> 
> Currently we are replacing the old model `CoverageResult` with the new model `CoverageNode` step by step. I think the model of the new delta computation (that is work in progress) will handle the problem you noticed automatically.



> [250..L259](https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java#L250..L259)
Without changing `CoverageResult`  won't work...  

That only let the option of the new model, `CoverageNode` anyway that I can help?

uhafner:
How are you using the delta computation right now? Are you using thresholds? Or just the fail options? Maybe it helps if you share your step call.

Can't we compute the delta from the serialized integer based values? That means we do not loose any precision in the serialization, just while computing the delta we can use doubles or BigDecimals. (I'm not sure if it makes sense to use a precision with more than 2 or 3 digits after the point.

uhafner:
I started jenkinsci/code-coverage-api-plugin#293 to improve the precision of the delta computation. It still needs to be integrated into the quality gates, so please be patient...

PR #324, opened 2022-02-21
Additional visualizations for the coverage report

fo-code:
This change is not yet ready to merge, but I would like to have some general feedback about the made changes before I finish the following TODOs:
- [x] Fix failing tests
- [x] Add tests for new code
- [x] Fix and adjust UI tests
- [x] Add JavaDoc
- [x] Integration test

Generally, I added additional visualizations (Treemap and Listview) of coverage information to the coverage report:
- Change Coverage (the coverage of only changed lines)
- Unexpected Coverage Changes (coverage changes in files without any edits)

I also changed the layout of the coverage report and made it possible to hide the overview charts in order to use to available space for analyzing the Treemap:
report_with_overview
report_without_overview

Also, I reworked the details view including the highlighted source code view:
report_details

Then, I adjusted to coverage summary in order to show only the most important coverage metrics line- and branch-coverage (the metric should become configurable in future releases).
summary

Last, I fixed some bugs and created some subdirectories in order to find required files faster:
- `mergePath` in `CoverageNode` was not able to handle files within the default package
- the coverage file painting created in `paint` in the class `SourceCodeFacade` started with offset 0, which caused the paints to be shifted by 1: 
paint_shift


plugin/src/main/java/io/jenkins/plugins/coverage/model/CodeDeltaCalculator.java
```diff
@@ -0,0 +1,111 @@
...
+        Map<String, CoverageNode> nodePathMapping = coverageNode.getAll(CoverageMetric.FILE).stream()
+                .map(node -> (FileCoverageNode) node)
+                .collect(Collectors.toMap(FileCoverageNode::getPath, Function.identity()));
+
+        List<FileChanges> fileChanges = delta.getFileChangesMap().values().stream()
+                .filter(fileChange -> fileChange.getFileEditType().equals(FileEditType.MODIFY)
+                        || fileChange.getFileEditType().equals(FileEditType.ADD))
+                .collect(Collectors.toList());
+
+        for (FileChanges fileChange : fileChanges) {
```
hd42, line 88: Warum erst in eine Liste packen wenn darüber dann nur iteriert wird? .forEach(fileChange->{...})

plugin/src/main/java/io/jenkins/plugins/coverage/model/CodeDeltaCalculator.java
```diff
@@ -0,0 +1,111 @@
...
+                    attachChangedCodeLines((FileCoverageNode) changedNode, fileChange);
+                }
+            }
+        }
+    }
+
+    private void attachChangedCodeLines(final FileCoverageNode changedNode, final FileChanges fileChange) {
+        Set<Change> relevantChanges = new HashSet<>();
+        relevantChanges.addAll(fileChange.getChangesByType(ChangeEditType.INSERT));
+        relevantChanges.addAll(fileChange.getChangesByType(ChangeEditType.REPLACE));
```
hd42, line 104: Da die changedCodeLines eh ein SortedSet sind braucht's hier noch kein Set um Duplikate auszufiltern, lieber die Schleife in eine eigene Methode und mit den Parametern INSERT/REPLACE aufrufen.

plugin/src/main/java/io/jenkins/plugins/coverage/model/CodeDeltaCalculator.java
```diff
@@ -0,0 +1,111 @@
...
+        for (Change change : relevantChanges) {
+            for (int i = change.getFromLine(); i <= change.getToLine(); i++) {
+                changedNode.addChangedCodeLines(i);
+            }
```
hd42, line 108: wenn addChangedCodeLines mit nur einer Zahl aufgerufen werden wäre eine addChangedCodeLine(int)-Methode besser, die dann nicht unnötig eine Zahl beim Aufruf in ein Array und eine List verpackt.

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageBuildAction.java
```diff
@@ -42,19 +44,25 @@
 public class CoverageBuildAction extends BuildAction<CoverageNode> implements HealthReportingAction, StaplerProxy {
     private static final long serialVersionUID = -6023811049340671399L;
 
-    static final String SMALL_ICON = "/plugin/code-coverage-api/icons/coverage.svg";
+    private static final CoverageTreeCreator COVERAGE_CALCULATOR = new CoverageTreeCreator();
+
+    public static final String SMALL_ICON = "/plugin/code-coverage-api/icons/coverage.svg";
     private static final String NO_REFERENCE_BUILD = "-";
```
hd42, line 50: Ich finde es unübersichtlich, wenn private und public Felder durchmischt werden. Üblich ist erst alle public Fields, dann alle private Fields, siehe z.B. https://checkstyle.sourceforge.io/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/DeclarationOrderCheck.html

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageNode.java
```diff
@@ -437,6 +463,30 @@ private CoverageNode createChild(final String childName) {
...
+        return copy;
+    }
+
+    protected void copyChildrenAndLeaves(final CoverageNode from, final CoverageNode to) {
+        to.getChildren().addAll(from.getChildren().stream()
+                .map(node -> node.copyTree(from))
+                .collect(Collectors.toList()));
+        to.getLeaves().addAll(from.getLeaves().stream()
+                .map(CoverageLeaf::copyLeaf)
+                .collect(Collectors.toList()));
```
hd42, line 487: Ich finde es unnötig Listen zu erstellen nur um dann addAll (welches meines Wissens auch nur wieder iteriert) aufzurufen.
from.getLeaves().stream()
                .map(CoverageLeaf::copyLeaf)
                .forEach(leaves::add)
spart diesen Zwischenschritt

plugin/src/main/java/io/jenkins/plugins/coverage/model/CodeDeltaCalculator.java
```diff
@@ -0,0 +1,130 @@
...
+
+import edu.hm.hafner.util.FilteredLog;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+import hudson.FilePath;
+import hudson.model.AbstractBuild;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.plugins.git.GitSCM;
+import hudson.plugins.git.util.BuildData;
```
uhafner, line 19: The code coverage plugin should have no dependency to Git!

plugin/src/main/java/io/jenkins/plugins/coverage/model/CodeDeltaCalculator.java
```diff
@@ -0,0 +1,130 @@
...
+                .collect(Collectors.toMap(
+                        fileChange -> getFullyQualifiedFileName(fileChange.getFileName()), Function.identity()));
+    }
+
+    /**
+     * Creates a {@link DeltaCalculator code delta calculator} which is compatible with the used SCM.
+     *
+     * @return an instance of the delta calculator
+     */
+    private DeltaCalculator createDeltaCalculator() {
```
uhafner, line 103: This fails in my jobs:

```
15:51:15  java.lang.ClassCastException: class org.jenkinsci.plugins.workflow.job.WorkflowRun cannot be cast to class hudson.model.AbstractBuild (org.jenkinsci.plugins.workflow.job.WorkflowRun is in unnamed module of loader jenkins.util.AntClassLoader @47695e1b; hudson.model.AbstractBuild is in unnamed module of loader org.eclipse.jetty.webapp.WebAppClassLoader @21b2e768)
15:51:15  	at io.jenkins.plugins.coverage.model.CodeDeltaCalculator.createDeltaCalculator(CodeDeltaCalculator.java:105)
15:51:15  	at io.jenkins.plugins.coverage.model.CodeDeltaCalculator.calculateCodeDeltaToReference(CodeDeltaCalculator.java:76)
15:51:15  	at io.jenkins.plugins.coverage.model.CoverageReporter.run(CoverageReporter.java:78)
15:51:15  	at io.jenkins.plugins.coverage.CoverageProcessor.performCoverageReport(CoverageProcessor.java:138)
15:51:15  	at io.jenkins.plugins.coverage.CoveragePublisher.perform(CoveragePublisher.java:111)
15:51:15  	at jenkins.tasks.SimpleBuildStep.perform(SimpleBuildStep.java:123)
15:51:15  	at org.jenkinsci.plugins.workflow.steps.CoreStep$Execution.run(CoreStep.java:101)
15:51:15  	at org.jenkinsci.plugins.workflow.steps.CoreStep$Execution.run(CoreStep.java:71)
15:51:15  	at org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution.lambda$start$0(SynchronousNonBlockingStepExecution.java:47)
15:51:15  	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
15:51:15  	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
15:51:15  	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
15:51:15  	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
15:51:15  	at java.base/java.lang.Thread.run(Unknown Source)
15:51:15  Finished: FAILURE
```

plugin/src/main/java/io/jenkins/plugins/coverage/model/CodeDeltaCalculator.java
```diff
@@ -0,0 +1,130 @@
...
+import io.jenkins.plugins.forensics.delta.DeltaCalculator;
+import io.jenkins.plugins.forensics.delta.DeltaCalculatorFactory;
+import io.jenkins.plugins.forensics.delta.model.Delta;
+import io.jenkins.plugins.forensics.delta.model.FileChanges;
+import io.jenkins.plugins.forensics.delta.model.FileEditType;
+import io.jenkins.plugins.forensics.git.delta.GitDeltaCalculatorFactory;
+
+/**
+ * Calculates the code delta between a Jenkins build and a reference build.
+ *
```
uhafner, line 31: Can't we obtain all SCM related information from the forensics-api plugin?

plugin/src/main/java/io/jenkins/plugins/coverage/model/visualization/code/SourceCodeProperties.java
```diff
@@ -0,0 +1,49 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+import io.jenkins.plugins.prism.SourceCodeRetention;
+
+/**
+ * Wraps the properties which are required for processing source code files.
+ *
+ * @author Florian Orendi
+ */
+public class SourceCodeProperties {
```
uhafner, line 14: Why is this class required? It does nothing?

uhafner (commented):
First part of my review. For the UI part I need to get a build with results first :-)

uhafner:
I created a new pipeline that breaks in build #82 with:

```
23:06:27  java.lang.NullPointerException
23:06:27  	at io.jenkins.plugins.coverage.model.coverage.FileCoverageDeltaProcessor.lambda$attachIndirectCoveragesChanges$2(FileCoverageDeltaProcessor.java:108)
23:06:27  	at java.base/java.util.TreeMap.forEach(Unknown Source)
23:06:27  	at io.jenkins.plugins.coverage.model.coverage.FileCoverageDeltaProcessor.attachIndirectCoveragesChanges(FileCoverageDeltaProcessor.java:104)
23:06:27  	at io.jenkins.plugins.coverage.model.CoverageReporter.run(CoverageReporter.java:99)
23:06:27  	at io.jenkins.plugins.coverage.CoverageProcessor.performCoverageReport(CoverageProcessor.java:137)
23:06:27  	at io.jenkins.plugins.coverage.CoveragePublisher.perform(CoveragePublisher.java:114)
23:06:27  	at jenkins.tasks.SimpleBuildStep.perform(SimpleBuildStep.java:123)
23:06:27  	at org.jenkinsci.plugins.workflow.steps.CoreStep$Execution.run(CoreStep.java:101)
23:06:27  	at org.jenkinsci.plugins.workflow.steps.CoreStep$Execution.run(CoreStep.java:71)
23:06:27  	at org.jenkinsci.plugins.workflow.steps.SynchronousNonBlockingStepExecution.lambda$start$0(SynchronousNonBlockingStepExecution.java:47)
23:06:27  	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)
23:06:27  	at java.base/java.util.concurrent.FutureTask.run(Unknown Source)
23:06:27  	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
23:06:27  	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
23:06:27  	at java.base/java.lang.Thread.run(Unknown Source)
```

You can reproduce by creating a new pipeline with the script:

```
node {
    stage('Checkout') {
        def url = "https://github.com/jenkinsci/analysis-model.git"
        env.COUNT = sh(script: "git ls-remote -t --refs $url | cut -f1 | wc -l", returnStdout: true).trim()
        if (env.COUNT.toInteger() > env.BUILD_NUMBER.toInteger()) {
            echo "Build: ${BUILD_NUMBER}"
            env.HASH = sh(script: "git ls-remote -t --refs $url | cut -f1 | head -n ${BUILD_NUMBER} | tail -1", returnStdout: true).trim()
            echo "Hash: ${HASH}"
            checkout([$class           : 'GitSCM',
                      branches         : [[name: "${HASH}"]],
                      userRemoteConfigs: [[url: "$url"]]])
            withMaven(maven: 'mvn-default', mavenLocalRepo: '/var/data/m2repository', mavenOpts: '-Xmx768m -Xms512m') {
                sh "mvn -V -U --fail-never -e jacoco:prepare-agent test jacoco:report -Djenkins.test.timeout=240 -Dmaven.test.failure.ignore"
            }
      gitDiffStat()
      publishCoverage(adapters: [jacocoAdapter('target/site/jacoco/jacoco.xml')], sourceFileResolver: sourceFiles('STORE_LAST_BUILD'))

      build(job: env.JOB_NAME, wait: false)
        } else {
            echo "Stopping after ${COUNT} tags"
        }
    }
}

```

This pipeline creates a new Jenkins build for each tag of the analysis model project. I used it to test the delta Forensics API but this works quite well for the coverage data. (It requires a night until all builds are finished 😉)

uhafner:
Some usability comments for the details: 
- We should detect if the current build has no sources attached at all: then the source code view should not be visible and the table should expand to the right. 
- If there is no source code for a file and the source view is visible then the text n/a *in the table* should be replaced by a *text without table* (the same style as in *No source code is selected.*, e.g. use the text *No source code available.*) 
- Sometimes you cannot open the details view, it simply ignores my click (build 79 works, build 80 not)
- Would it make sense to show the project coverage of a file as additional column when showing the delta coverage? 
- We should store the selected UI setup in the browser storage
- It would be helpful to see in a full glance the numbers for project vs. delta coverage. Maybe it makes sense to show the coverage overview for the project on the top and for the change at the bottom? Then we can remove the trend chart. 
- If the layout is split into left and right then the boxes should span the same vertical space:
Bildschirmfoto 2022-03-16 um 09 33 10
- somehow the padding of the source code table is broken now, there is no gap between table and pagination anymore:
New:
Bildschirmfoto 2022-03-16 um 09 13 59
Before:
Bildschirmfoto 2022-03-16 um 09 14 21
- Indirect coverage can be empty:
Bildschirmfoto 2022-03-16 um 09 36 27

uhafner:
The overview looks really nice now!

Bildschirmfoto 2022-03-16 um 09 44 19

Would it make sense to add a delta for the change coverage as well (with respect to the project coverage)?

plugin/src/main/java/io/jenkins/plugins/coverage/model/FileCoverageNode.java
```diff
@@ -16,6 +21,11 @@
 
     private final String sourcePath;
 
+    // new since 3.0.0
+    private SortedMap<Integer, Coverage> coveragePerLine = new TreeMap<>();
```
uhafner, line 25: Shouldn't we store the type as well? I.e. for line 4 we have in the JaCoCo report: 
```
<line nr="4" mi="2" ci="14" mb="1" cb="5"/>
```

This should produce a 
- Line Coverage of 1 / 1
- Instruction Coverage of 14 / 16
- Branch Coverage of 5 / 6

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageNode.java
```diff
@@ -319,6 +358,22 @@ public Coverage getCoverage(final CoverageMetric searchMetric) {
         return childNodes;
     }
 
+    /**
+     * Returns recursively all nodes of the instance {@link FileCoverageNode}.
+     *
+     * @return all file coverage nodes
+     */
+    public List<FileCoverageNode> getAllFileCoverageNodes() {
```
uhafner, line 366: Is this method required? Wouldn't it be simpler to use `getAll` and work on the `CoverageNode` types?

plugin/src/main/java/io/jenkins/plugins/coverage/model/FileCoverageNode.java
fo-code: At the moment, the stored `Coverage` provides information about line- and branch-coverage. In my opinion, instruction-coverage is not required here, since I guess in the most cases, it is not used.
Anyways, I would like to refactor this at all and store the `coverageLineMapping` within `MethodCoverageNode` since at the moment, the coverage mapping is not applyable on methods. The required storage would be the same and at the same time, we can use the lines numbers per method within the mapping for filtering source code for specific methods in further releases.
If instruction-coverage is required and worth the additional storage, I can wrap the `Coverage` together with its metric when I apply the described refactoring. 
What is your opinion on this?

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageNode.java
fo-code: I wrote this method since otherwise, there has to be a type check everytime it is used. Otherwise, there will be cast exception when for example the reference build has been built with an older release, where `FileCoverageNode` did not exist and `getAll` returns nodes of the type `CoverageNode` only.

uhafner: But why do you need to cast anyway? You are using only methods of `CoverageNode`.

fo-code:
@uhafner thank you very much for the feedback!
Locally, I already reworked the code based on your feedback - I will push the changes soon.
Some UI adjustments:
- the builld result overview now shows also the percentage of covered vs. missed coverage changes
build_overview
- the coverage table now shows the file coverage delta against the file coverage in the reference build. Also, I adopted the design from the dashboard column since it saves some space.
table_pc
- the change coverage table uses the described coverage delta field for displaying the delta between the total file coverage and the change coverage within this file (here, the change branch-coverage is 75%, which is 8,33% higher then the total file coverage)
table_cc

uhafner:
This looks awesome! 

> the builld result overview now shows also the percentage of covered vs. missed coverage changes

Shouldn't the delta of the line coverage be -36.73% in your example screenshot? 

> the coverage table now shows the file coverage delta against the file coverage in the reference build. Also, I adopted the design from the dashboard column since it saves some space.

The new visualization looks much better! Though some cells correctly render the percentage as width, and some do not. Is this intended? Main.java color bar width looks correct for line coverage, but not for line coverage delta?

Can you try to shorten the column names and tab names as much as possible? Since we are showing the coverage results we can simplify the naming.  I made a demo using my laptop and this rendered several texts into multiple lines. So for the tabs use "Project", "Change" and "Indirect Change". The word "Coverage" can be omitted. For the columns it would make sense to just use "Line", "Branch", and "Line Δ" or "Line 𝚫".

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageNode.java
fo-code: I use this method in classes like `FileCoverageDeltaProcessor` where I got the root of the tree as `CoverageNode` and I need all `FileCoverageNode` of the tree since only this subclass stores the information about changed lines or the coverage per line - here, I would have to cast to access these information.
For example, when the reference root has been built with an older version, this cast would fail if I do not check it first. Since there are many cases like this, I wrote this method `getAllFileCoverageNodes`.

fo-code:
> Shouldn't the delta of the line coverage be -36.73% in your example screenshot?

Oh yes that is confusing. I took the screenshots from different builds since my focus was to show the visualization.

> Can you try to shorten the column names

Shorten the names is a good idea - I will do that.

> Though some cells correctly render the percentage as width, and some do not. Is this intended?

The width of the colorization is as it is indended. It is done like for the dashboard column, where the delta is colorized dependent on the coverage tendency (increase, decrease, equals). 
In my opinion, here it is only relevant to find out whether the coverage increased or decreased. Since for deltas there are often small values, this might also look misleading when colorized like the coverage and even increases are red. 
These are my thoughts behind the colorization.

uhafner:
> > Though some cells correctly render the percentage as width, and some do not. Is this intended?
> 
> The width of the colorization is as it is indended. It is done like for the dashboard column, where the delta is colorized dependent on the coverage tendency (increase, decrease, equals). In my opinion, here it is only relevant to find out whether the coverage increased or decreased. Since for deltas there are often small values, this might also look misleading when colorized like the coverage and even increases are red. These are my thoughts behind the colorization.

I see. Maybe it makes sense to tweak the rendering somewhat to make it distinguishable from the other column rendering, otherwise it might confuse users. Maybe we can work with borders or use a badge style without filling the whole cell. I'll experiment a little bit once your changes are pushed...

plugin/src/main/java/io/jenkins/plugins/coverage/model/FileCoverageNode.java
uhafner: No this is not required. I don't use instruction coverage on my own. So I am fine with just recording line and branch coverage. But even if we do not use instruction coverage there should be values for line *and* branch coverage. Or are you storing one tree with line coverage and one for branch coverage?

fo-code:
@uhafner this would be my proposal for the design of the table delta columns - your idea with the badges looks good in my opinion:
table_badges
What do you think about it?

I figured out that there is a bug with the data table options when one column is hidden. We can talk about this in the next meeting.
After that, I also would like to adjust the widths of the columns in order to make better use of the space - especially on small screens.

plugin/src/main/java/io/jenkins/plugins/coverage/model/FileCoverageNode.java
fo-code: No, there is just one `Coverage` for each line. The information "line or branch" is given automatically by `getTotal()`, since line coverage has the value `1` and branch coverage has `>1`.
The logic I wrote for calculating coverage deltas distinguishes this already.
The reason I decided this is that having two `Coverage` per line would double the required space, which means that for a project with 100.000 LOC the XML would be nearly 100.000 Coverage-Tags larger.

uhafner:
> @uhafner this would be my proposal for the design of the table delta columns - your idea with the badges looks good in my opinion: table_badges What do you think about it?
> 
> I figured out that there is a bug with the data table options when one column is hidden. We can talk about this in the next meeting. After that, I also would like to adjust the widths of the columns in order to make better use of the space - especially on small screens.

Yes, that looks much better!

plugin/src/main/java/io/jenkins/plugins/coverage/model/FileCoverageNode.java
uhafner: I see! Then please add a small JavaDoc comment for that field that describes that modeling.

plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageViewModel/index.jelly
```diff
@@ -5,8 +5,8 @@
 
   <bs:page it="${it}" notitle="true" class="fluid-container d-flex flex-column h-100">
 
-    <st:adjunct includes="io.jenkins.plugins.jquery3"/>
     <st:adjunct includes="io.jenkins.plugins.echarts"/>
+    <st:adjunct includes="io.jenkins.plugins.data-tables-select"/>
```
uhafner, line 9: No the selection works in the table!

🎉

fo-code:
I added a modified source code highlighting, that only affected lines are highlighted and visible: 
Change coverage:
sourcecode_cc
Indirect coverage changes:
sourcecode_icc

uhafner:
I started to improve the memory footprint (XML and runtime), see #374 for details.

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageBuildAction.java
```diff
@@ -209,6 +312,153 @@ public String formatDelta(final CoverageMetric metric) {
...
+     * @return the change coverage metric
+     */
+    @SuppressWarnings("unused") // Called by jelly view
+    public String formatChangeCoverage(final CoverageMetric metric) {
+        String coverage = Messages.Coverage_Not_Available();
+        if (changeCoverage != null && changeCoverage.containsKey(metric)) {
+            coverage = FractionFormatter
+                    .formatFraction(changeCoverage.get(metric), Functions.getCurrentLocale());
+        }
+        return metric.getName() + ": " + coverage;
```
uhafner, line 331: Duplicate to lines 345-350: extract into a method and use the map as parameter.

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageJobAction.java
```diff
@@ -47,7 +48,7 @@ public String getSearchUrl() {
     }
 
     @Override
-    protected LinesChartModel createChartModel(final String configuration) {
+    public LinesChartModel createChartModel(final String configuration) {
```
uhafner, line 51: Is this still required?

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageNode.java
```diff
@@ -56,11 +62,48 @@ public class CoverageNode implements Serializable {
...
+     */
+    @VisibleForTesting
+    public CoverageNode(final CoverageMetric metric, final String name, final CoverageTreeCreator coverageTreeCreator) {
         this.metric = metric;
         this.name = name;
+        this.coverageTreeCreator = coverageTreeCreator;
+    }
+
+    /**
+     * Called after de-serialization to retain backward compatibility.
```
uhafner, line 86:
```suggestion
     * Called after de-serialization to restore transient fields.
```

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageViewModel.java
```diff
@@ -106,43 +127,48 @@ public TreeMapNode getCoverageTree() {
...
-        return JACKSON_FACADE.toJson(createTrendChart(configuration));
+    public TreeMapNode getCoverageChangesTree() {
+        return TREE_MAP_NODE_CONVERTER.toTeeChartModel(indirectCoverageChangesTreeRoot);
     }
 
-    private Optional<CoverageBuildAction> getLatestAction() {
-        return Optional.ofNullable(getOwner().getAction(CoverageBuildAction.class));
+    /**
+     * Returns the table model that shows the files along with the branch and line coverage. Currently, only one table
+     * is shown in the view, so the ID is not used.
```
uhafner, line 155: This is not true anymore

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageViewModel.java
```diff
@@ -293,34 +400,65 @@ private SortedMap<CoverageMetric, Coverage> getMetricsDistribution() {
...
             List<TableColumn> columns = new ArrayList<>();
 
-            columns.add(new TableColumn("Package", "packageName"));
-            columns.add(new TableColumn("File", "fileName"));
-            columns.add(new TableColumn("Line Coverage", "lineCoverageValue").setHeaderClass(ColumnCss.PERCENTAGE));
-            columns.add(new TableColumn("Line Coverage", "lineCoverageChart", "number"));
-            columns.add(new TableColumn("Branch Coverage", "branchCoverageValue").setHeaderClass(ColumnCss.PERCENTAGE));
-            columns.add(new TableColumn("Branch Coverage", "branchCoverageChart", "number"));
+            // TODO: DataTables API sets options wrong in case of invisible columns (columnDef and <th> number not matching)
+            // TODO: DataTables API requires more granular width options
```
uhafner, line 431: Is there anything I can do here in the data tables API?

plugin/src/main/java/io/jenkins/plugins/coverage/model/MethodCoverageNode.java
```diff
@@ -0,0 +1,78 @@
...
+    protected MethodCoverageNode copyTree(@CheckForNull final CoverageNode copiedParent) {
+        MethodCoverageNode copy = new MethodCoverageNode(getName(), getLineNumber());
+        if (copiedParent != null) {
+            copy.setParent(copiedParent);
+        }
+
+        copyChildrenAndLeaves(this, copy);
+
+        return copy;
+    }
```
uhafner, line 57: This code is duplicated in all classes an can be simplified using a factory method (see https://github.com/uhafner/coverage-model/blob/main/src/main/java/edu/hm/hafner/coverage/CoverageNode.java#L463)

plugin/src/main/java/io/jenkins/plugins/coverage/model/util/FractionFormatter.java
```diff
@@ -30,6 +30,20 @@ public static Fraction transformFractionToPercentage(final Fraction fraction) {
         return fraction.multiplyBy(HUNDRED);
     }
 
+    /**
```
uhafner, line 33: Currently it is somewhat unclear throughout the whole plugin whether a `Fraction` instance ist used as a rational number or a percentage. Maybe it would make sense to create a new `Percentage` class that has a `Fraction` (multiplied by 100) as a field. Does that make more sense?

plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageBuildAction/summary.jelly
```diff
@@ -4,38 +4,103 @@
...
+          </j:otherwise>
+        </j:choose>
+        <j:if test="${it.hasIndirectCoverageChanges()}">
+          <li>
+            <b>
+              <a href="coverage/#coverageChangesTree">Indirect Coverage Changes:</a>
+            </b>
+            <ul>
+              <j:forEach var="metric" items="${metrics}">
+                <span id="indirect-coverage-changes"
```
uhafner, line 88: Can you use the class `jenkins-tooltip` so that the tooltip uses the new tooltip styles? Best thing would be to install https://plugins.jenkins.io/design-library/ for reference.

uhafner (changes requested):
Just a couple of small things still need to be done. The overall result already is really impressive!

plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageBuildAction/summary.jelly
timja: viewable online at https://weekly.ci.jenkins.io/design-library/Tooltips/

fo-code:
Thanks for the comments! I will finalize the code and add an integration test next week.
Then, the PR should be ready to be merged.

plugin/src/main/java/io/jenkins/plugins/coverage/model/util/FractionFormatter.java
fo-code: I added a class `CoveragePercentage` which can be used for displaying and serializing coverage values as percentage.
`Fraction` is still in use for calculations since it can represent coveraged and missed lines properly and also provides math operations.
Now, `Fraction` is not used for displaying or serializing coverage anymore and it is clear whether a fraction or a percentage is represented by a value.

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageViewModel.java
fo-code: I will show you what I mean in the next meeting.

plugin/src/test/java/io/jenkins/plugins/coverage/model/DockerGitForensicsITest.java
```diff
@@ -0,0 +1,294 @@
...
+import static org.assertj.core.api.Assertions.*;
+import static org.assertj.core.api.Assumptions.*;
+
+/**
+ * Tests the integration of the Forensics API Plugin while using its Git implementation.
+ *
+ * @author Florian Orendi
+ */
+@Testcontainers(disabledWithoutDocker = true)
+class DockerGitForensicsITest extends IntegrationTestWithJenkinsPerSuite {
```
uhafner, line 38: I think it is sufficient to use `GitForensicsITest` as there is no other implementation

plugin/src/test/java/io/jenkins/plugins/coverage/model/DockerGitForensicsITest.java
```diff
@@ -0,0 +1,294 @@
...
+    /**
+     * Verifies the calculated overall coverage including the coverage delta.
+     *
+     * @param action
+     *         The created Jenkins action
+     */
+    private void verifyOverallCoverage(final CoverageBuildAction action) {
+        assertThat(action.getLineCoverage()).satisfies(coverage -> {
+            assertThat(coverage.getCovered()).isEqualTo(546);
+            assertThat(coverage.getMissed()).isEqualTo(461);
```
uhafner, line 156: Would be more clean to use the generated assertions:
```
assertThat(action.getLineCoverage()).satisfies(coverage -> {
            assertThat(coverage).hasCovered(546).hasMissed(461);
        });
```

Or delegate to equals:

```
        assertThat(action.getLineCoverage()).isEqualTo(new Coverage(546, 461));
```

plugin/src/test/java/io/jenkins/plugins/coverage/model/DockerGitForensicsITest.java
```diff
@@ -0,0 +1,294 @@
...
+        });
+        assertThat(action.getBranchCoverage()).satisfies(coverage -> {
+            assertThat(coverage.getCovered()).isEqualTo(136);
+            assertThat(coverage.getMissed()).isEqualTo(94);
+        });
+        assertThat(action.getDifference()).contains(
+                new SimpleEntry<>(LINE, CoveragePercentage.getCoveragePercentage(65_160, 103_721)),
```
uhafner, line 163: maybe rename the simple factory method to `valueOf`?
```suggestion
                new SimpleEntry<>(LINE, CoveragePercentage.valueOf(65_160, 103_721)),
```

plugin/src/test/java/io/jenkins/plugins/coverage/model/DockerGitForensicsITest.java
```diff
@@ -0,0 +1,294 @@
...
+        assertThat(root).isNotNull();
+
+        List<FileCoverageNode> changedFiles = root.getAllFileCoverageNodes().stream()
+                .filter(fileNode -> !fileNode.getChangedCodeLines().isEmpty())
+                .collect(Collectors.toList());
+        assertThat(changedFiles).hasSize(4);
+        assertThat(changedFiles.stream()
+                .map(FileCoverageNode::getChangedCodeLines)
+                .flatMap(Collection::stream))
+                .containsExactlyInAnyOrder(15, 17, 63, 68, 80, 90, 130);
```
uhafner, line 221: AssertJ has native support for such things:

```
        assertThat(changedFiles).extracting(FileCoverageNode::getChangedCodeLines)
                .containsExactly(StreamEx.of(15, 17, 63, 68, 80, 90, 130).toCollection(TreeSet::new));

```

plugin/src/test/java/io/jenkins/plugins/coverage/model/DockerGitForensicsITest.java
```diff
@@ -0,0 +1,294 @@
...
+        assertThat(changedFiles.stream()
+                .map(FileCoverageNode::getName))
+                .containsExactlyInAnyOrder("MinerFactory.java", "RepositoryMinerStep.java",
+                        "SimpleReferenceRecorder.java", "CommitDecoratorFactory.java");
+
+        assertThat(root.getFileAmountWithChangedCoverage()).isEqualTo(2);
```
uhafner, line 227:
```suggestion
        assertThat(root).hasFileAmountWithChangedCoverage(2);
```

plugin/src/test/java/io/jenkins/plugins/coverage/model/CoverageBuildActionTest.java
```diff
@@ -57,22 +69,154 @@ void shouldGetCoverageDifferenceForSpecifiedMetric() {
...
+                .isEqualTo(COVERAGE_PERCENTAGE.formatDeltaPercentage(LOCALE));
+    }
+
+    @Test
+    void shouldGetChangeCoverageDifferences() {
+        CoverageBuildAction action = createCoverageBuildActionWithMocks();
+        assertThat(action.hasChangeCoverageDifference(COVERAGE_METRIC)).isTrue();
+        assertThat(action.getChangeCoverageDifference(COVERAGE_METRIC))
+                .isNotNull()
+                .satisfies(coverage -> assertThat(coverage).isEqualTo(COVERAGE_PERCENTAGE));
```
uhafner, line 81:
```suggestion
        assertThat(action.getChangeCoverageDifference(COVERAGE_METRIC)).isEqualTo(COVERAGE_PERCENTAGE));
```

plugin/src/test/java/io/jenkins/plugins/coverage/model/CoverageBuildActionTest.java
```diff
@@ -57,22 +69,154 @@ void shouldGetCoverageDifferenceForSpecifiedMetric() {
...
+    @Test
+    void shouldGetChangeCoverageForSpecifiedMetric() {
+        CoverageBuildAction action = createChangeCoverageBuildActionWithMocks();
+        assertThat(action.hasChangeCoverage()).isTrue();
+        assertThat(action.hasCodeChanges()).isTrue();
+        assertThat(action.hasChangeCoverage(COVERAGE_METRIC)).isTrue();
+        assertThat(action.getChangeCoverage(COVERAGE_METRIC))
+                .isNotNull()
+                .satisfies(coverage -> assertThat(coverage.getCoveredPercentage()).isEqualTo(COVERAGE_PERCENTAGE));
+    }
```
uhafner, line 93:
```suggestion
        assertThat(action.getChangeCoverage(COVERAGE_METRIC))
                .extracting(Coverage::getCoveredPercentage)
                .isEqualTo(COVERAGE_PERCENTAGE);
```

uhafner (approved):
The tests look good! Some small comments about AssertJ usage though.

Another thing: would it be possible to add some assertions for the new code in the existing test `DeltaComputationVsReferenceBuildITest`? Even if the new code is not used at all it makes sense to see what values are returned in such a case.

fo-code:
> The tests look good! Some small comments about AssertJ usage though.

Thanks for the hints, I fixed this.

> Another thing: would it be possible to add some assertions for the new code in the existing test `DeltaComputationVsReferenceBuildITest`? Even if the new code is not used at all it makes sense to see what values are returned in such a case.

Yes, I added some assertions regarding the calculated code delta and the change coverage.
I am not sure why the old model has been tested here, but I refactored the test class and fixed that one existing bug.
Is that okay?

uhafner:
> > The tests look good! Some small comments about AssertJ usage though.
> 
> Thanks for the hints, I fixed this.
> 
> > Another thing: would it be possible to add some assertions for the new code in the existing test `DeltaComputationVsReferenceBuildITest`? Even if the new code is not used at all it makes sense to see what values are returned in such a case.
> 
> Yes, I added some assertions regarding the calculated code delta and the change coverage. I am not sure why the old model has been tested here, but I refactored the test class and fixed that one existing bug. Is that okay?

Yes that is fine. (The old code will be removed step by step.)

PR #372, opened 2022-04-22
If the coverage decreases on a PR, mark the check as failed

### LINKED ISSUE #256: Publish failed check for coverage drop
### Feature Request

There doesn't seem to be any way to configure coverage to publish a failed check if coverage drops vs the reference build.. The entire build can be failed, but if the only reason for that failure is a drop in coverage, the check won't reflect that - it will show as a success because there's no code path that sets a failure message if the coverage diff is negative.

It would be good if the published check would reflect a failure in the check if the build is failed because of it.
It would be _great_ if it could report the drop in the check (or be configured to report it) even if failBuildIfCoverageDecreasedInChangeRequest is set false - that way the check could still be used as a warning flag without necessarily failing a build.

garethjevans:
When `failBuildIfCoverageDecreasedInChangeRequest` is set, as well as failing the build by throwing the exeception this PR adds a failMessage to the action to ensure that when the check is published it is marked as failed.

Fixes #256

### DIFF
--- a/plugin/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
+++ b/plugin/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
@@ -21,7 +21,6 @@ import java.util.Set;
 import java.util.TreeMap;
 import java.util.stream.Collectors;
 
-import io.jenkins.plugins.coverage.model.Coverage;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.lang.StringUtils;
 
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -247,12 +248,15 @@ private Optional<CoverageAction> getPreviousResult(final Run<?, ?> startSearch)
...
+    private void failBuildIfChangeRequestDecreasedCoverage(final CoverageResult coverageResult, final CoverageAction action)
             throws CoverageException {
         float coverageDiff = coverageResult.getCoverageDelta(CoverageElement.LINE);
         if (coverageDiff < 0) {
-            throw new CoverageException(
-                    "Fail build because this change request decreases line coverage by " + coverageDiff);
+            String message = "Fail build because this change request decreases line coverage by " + coverageDiff;
+            if (StringUtils.isBlank(action.getFailMessage())) {
+                action.setFailMessage(message);
+            }
```
uhafner, line 258: Can you make that more object oriented by moving that code to the action?

uhafner (changes requested):

uhafner (approved):

PR #374, opened 2022-04-23
Improve memory footprint

uhafner:
Improve XStream serialization and memory footprint for coverage instances:
- [X] Use alias for each type of coverage node.
- [X] Use custom converter for coverage values.
- [X] Use custom converter for coverage percentages
- [X] Use custom converter for coverage per line number mapping
- [x] Use custom converter for delta coverage 
- [X] Create `Coverage` instances using a builder
- [X] Use Flyweight pattern for `Coverage` instances

The analysis-model XML now is 3.7M rather than 5.7M before. 

Bildschirmfoto 2022-05-16 um 23 43 09

After compacting the new mappings of line to coverage the size is even smaller (2.7M):

Bildschirmfoto 2022-05-18 um 09 45 32

In a final step, all 4 new maps and sets have been simplified and we now have 2.3M:

Bildschirmfoto 2022-05-20 um 22 14 22


plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageBuildAction.java
```diff
@@ -556,10 +580,15 @@ static class CoverageXmlStream extends AbstractXmlStream<CoverageNode> {
         @Override
         protected void configureXStream(final XStream2 xStream) {
             xStream.alias("node", CoverageNode.class);
+            xStream.alias("package", PackageCoverageNode.class);
+            xStream.alias("file", FileCoverageNode.class);
+            xStream.alias("method", MethodCoverageNode.class);
             xStream.alias("leaf", CoverageLeaf.class);
             xStream.alias("coverage", Coverage.class);
+            xStream.alias("fraction", Fraction.class);
```
fo-code, line 588: `Fraction` is not used for serialization anymore. This alias can be removed.

fo-code (changes requested):
Looks goot to me! Just one little thing...

fo-code:
> Maybe we should serialize to something like:
> 
> ```
> <coveragePerLine>33: 1/1, 34: 1/1, 35: 1/1, 48: 1/2, ...</coveragePerLine>
> ```

@uhafner Yes, the coverage/line-mapping is the place where a lot of storage can be saved - the other maps used in `FileCoverageNode` are not relevant for this.
I calculated the storage usage: 
This mapping causes 38% of the required storage for the single node representation of a file with 87 lines which have coverage information.

Your idea for the serialization looks like the perfect improvement for this in my opinion.

fo-code (approved):
Looks good to me! I tested it locally and the required disk space decreased a lot.

PR #394, opened 2022-05-20
Improve the layout and design of the coverage columns

uhafner:
## Main table 

Bildschirmfoto 2022-05-20 um 16 30 18

## Coverage table

Bildschirmfoto 2022-05-20 um 17 16 26

### DIFF
--- a/plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageViewModel.java
+++ b/plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageViewModel.java
@@ -582,13 +582,13 @@ public class CoverageViewModel extends DefaultAsyncTableContentProvider implemen
             double percentage = coverage.getDoubleValue();
             String sort = String.valueOf(percentage);
             DisplayColors colors = CoverageLevel.getDisplayColorsOfCoverageLevel(percentage, COLOR_PROVIDER);
-            String tag = span()
-                    .withTitle(tooltip)
-                    .withStyle(String.format(
-                            "color:%s; background-image: linear-gradient(90deg, %s %f%%, transparent %f%%); display:block;",
+            String tag = div().withClasses("coverage-column-outer").with(
+                    div().withClasses("coverage-column-inner")
+                            .withStyle(String.format("color:%s; background-image: linear-gradient(90deg, %s %f%%, transparent %f%%);",
                             colors.getLineColorAsHex(), colors.getFillColorAsHex(),
                             percentage, percentage))
-                    .withText(text)
+                            .withTitle(tooltip)
+                            .withText(text))
                     .render();
             return new DetailedColumnDefinition(tag, sort);
         }
@@ -610,12 +610,12 @@ public class CoverageViewModel extends DefaultAsyncTableContentProvider implemen
             String sort = String.valueOf(coverageValue);
             DisplayColors colors = CoverageChangeTendency
                     .getDisplayColorsForTendency(coverageValue, COLOR_PROVIDER);
-            String tag = span()
-                    .withClasses("badge", "badge-delta")
+            String tag = div().withClasses("coverage-column-outer").with(
+                    div().withClasses("coverage-column-inner")
                     .withStyle(String.format("color:%s;background-color:%s;",
                             colors.getLineColorAsHex(), colors.getFillColorAsHex()))
                     .withText(coverageText)
-                    .withTitle(tooltip)
+                    .withTitle(tooltip))
                     .render();
             return new DetailedColumnDefinition(tag, sort);
         }
--- a/plugin/src/main/resources/io/jenkins/plugins/coverage/model/visualization/dashboard/CoverageColumn/column.jelly
+++ b/plugin/src/main/resources/io/jenkins/plugins/coverage/model/visualization/dashboard/CoverageColumn/column.jelly
@@ -1,27 +1,30 @@
 <?jelly escape-by-default='true'?>
 <j:jelly xmlns:j="jelly:core">
 
+  <link rel="stylesheet" href="${resURL}/plugin/code-coverage-api/css/column-style.css"/>
+
   <j:set var="coverageText" value="${it.getCoverageText(job)}"/>
   <j:set var="coverageValue" value="${it.getCoverageValue(job)}"/>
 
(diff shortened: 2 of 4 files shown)


plugin/src/main/resources/io/jenkins/plugins/coverage/model/visualization/dashboard/CoverageColumn/column.jelly
```diff
@@ -1,27 +1,30 @@
...
 
+  <link rel="stylesheet" href="${resURL}/plugin/code-coverage-api/css/column-style.css"/>
+
   <j:set var="coverageText" value="${it.getCoverageText(job)}"/>
   <j:set var="coverageValue" value="${it.getCoverageValue(job)}"/>
 
   <j:choose>
     <j:when test="${coverageValue.isPresent()}">
       <j:set var="displayColors" value="${it.getDisplayColors(job, coverageValue)}"/>
-      <j:set var="lineColor" value="${displayColors.getLineColorAsHex()}"/>
```
fo-code, line 10: The intention behind this line is that it is guaranteed that the color of the text is readable on the background color.
With the currently used color palette, this shouldn't be a problem.
I guess this has been removed to make all URLs in the dashboard the same color?
When this line is removed, we have to take care that there are no background colors in use which causes problems with the readability.

plugin/src/main/resources/io/jenkins/plugins/coverage/model/visualization/dashboard/CoverageColumn/column.jelly
```diff
@@ -1,27 +1,30 @@
...
-        </j:choose>
+      <td align="left">
+        <div class="coverage-column-outer">
+          <div class="coverage-column-inner" style="background-image: linear-gradient(90deg, ${backgroundColor} ${fillPercentage}, transparent ${fillPercentage});">
+            <j:choose>
+              <j:when test="${url.isEmpty()}">
+                ${coverageText}
+              </j:when>
+              <j:otherwise>
+                <a style="color:${lineColor}" href="${rootURL}/${job.url}lastSuccessfulBuild/${url}">${coverageText}</a>
```
fo-code, line 23: `style="color:${lineColor}"` can be removed if there should always be the default URL color in use.

fo-code (changes requested):
The new design looks a lot better!
Just one little thing that i noticed...

plugin/src/main/resources/io/jenkins/plugins/coverage/model/visualization/dashboard/CoverageColumn/column.jelly
uhafner: When we set a color then we should use the predefined color variables of Jenkins. Without the color, the links are using the Jenkins default, which should look good in all themes:

Bildschirmfoto 2022-05-22 um 22 11 24

plugin/src/main/resources/io/jenkins/plugins/coverage/model/visualization/dashboard/CoverageColumn/column.jelly
uhafner: That's right. I just deleted the variable and forgot to remove the reference.

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageViewModel.java
```diff
@@ -558,39 +558,32 @@ public DetailedColumnDefinition getLoc() {
...
-                    .render();
-            return new DetailedColumnDefinition(tag, sort);
+        protected DetailedColumnDefinition createColoredCoverageColumn(final Coverage coverage, final String tooltip) {
+            double percentage = coverage.getCoveredPercentage().getDoubleValue();
+            if (coverage.isSet()) {
+                DisplayColors colors = CoverageLevel.getDisplayColorsOfCoverageLevel(percentage, COLOR_PROVIDER);
+                String cell = div().withClasses(COVERAGE_COLUMN_OUTER).with(
+                                div().withClasses(COVERAGE_COLUMN_INNER)
+                                        .withStyle(String.format(
+                                                "color:%s; background-image: linear-gradient(90deg, %s %f%%, transparent %f%%);",
```
uhafner, line 578: I think here we have the same problem. The whole UI looks somewhat funny in dark mode (I need to update the datatables CSS as well)

Bildschirmfoto 2022-05-22 um 22 36 02

uhafner:
Yes, the dark mode needs to be fixed in an additional PR. 

PS: I did not merge the changes yet since the layout still needs to be improved for small screens. Currently the source code and the tables are only visible by 50% when used on a laptop. I started refactoring the data tables plugin so that superfluous columns will be hidden on small screens automatically. The works quite well already (not yet pushed). Hopefully I find some time next week to improve the source code view for small views as well. I think for those screens it might make sense to restore the old behavior of having a separate view for the source code and one for the table. Or do you have a better idea, @fo-code?

fo-code:
Yes you are right. 
This is a problem I already tried to fix in some way, using the width option for columns in `CoverageViewModel`.
That didn't work for the reason I mentioned in the TODOs there, but it can enhance the view for small screens.
I already tried that manuelly.
Anyway, I guess your approach sounds good since there might be additional columns in the future regarding complexity etc..

One option is that we could add an option to hide the integrated source code view, as already done for the coverage charts within the overview.
When the integrated source code is activated, everything works as it is right now.
Otherwise (when the integrated source code view is deactivated using the toggle button) the row selection calls the former method which opens a new tab - we can check this Javascript - and the table takes the whole screen.

Another option would be to always show the source code view next to the table, but put the focus on the table that it has always its required size - the source code is then only visible with horizont
Then, add links to the file names to provide the option to open the source code in a new tab, as it has be done bedore.

The most easy option is, that we change only the UI that the source code is shown below the table if the screen is to small to show both cards next to each other in the required size.
I my opinion, that has the same effect as opening a new tab, but it requires no changes of the logic, additional tests etc..

I think these are approaches which covers every use case and can be integrated quite easily, since the code for opening new tabs already exist.
I would prefer the last mentioned option, since this is already done like this for other parts in Jenkins - at least that's what I saw.
What do you think about that @uhafner?

PR #417, opened 2022-07-02
Add a coberturaReportAdapter symbol to disambiguate from Cobertura plugin's symbols

medianick:
Allows for this report adapter to be disambiguated from the one provided by the Cobertura plugin in the case both plugins are installed; see https://stackoverflow.com/a/72839475/466874 for this specific case and https://issues.jenkins.io/browse/INFRA-2012 and https://github.com/jenkins-infra/helpdesk/issues/1662 for the fundamental issue.


uhafner:
Thanks for fixing that. I totally forgot about the old cobertura plugin.

medianick:
> Thanks for fixing that. I totally forgot that the old cobertura plugin still is used.

Thanks for merging it! This will give us a smoother path toward removing the Cobertura plugin since we'll be able to switch to the new `coberturaReportAdapter` adapter symbol for our existing jobs first, then remove the plugin, then (if desired) switch to the `cobertura` adapter symbol without conflict/error.

PR #423, opened 2022-07-10
Use the coverages that are persisted within the action if possible

uhafner:
Otherwise, the whole coverage result XML will be loaded just for a single value.

Will fix jenkins-infra/helpdesk#3029.


medianick:
Thanks @uhafner! I look forward to trying this fix. As I noted on the linked issue, our PR pages for multibranch pipeline jobs started taking over a minute to render a list of 100 PRs. We hadn't yet tracked down the cause but it coincided with the initial upgrade to 3.0.0.

Issue #441, opened 2022-08-03
Painting does not clean up temporary directories


stefan6419846:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
  Jenkins: 2.361
  OS: Linux - 5.3.18-150300.59.71-default
  ---
  [...]
  code-coverage-api:3.0.4
  [...]
  ```

</details>

### What Operating System are you using (both controller, and any agents involved in the problem)?

openSUSE Leap 15.3

### Reproduction steps

1. Configure a coverage reporter for the build pipeline (Python using *coverage.py* in this case).
2. Wait for the coverage report to have been generated:

```
[Coverage] Reference build recorder is not configured
[Coverage] -> No reference build defined, falling back to previous build
[Coverage] -> Found no reference result in reference build
[Coverage] Obtaining action of reference build
[Coverage] Reference build recorder is not configured
[Coverage] -> Found no reference result in reference build
[Coverage] Executing source code painting...
[Coverage] Painting 701 source files on agent
[Coverage] Searching for source code files in:
[Coverage] -> /home/jenkins/.some-agent/workspace/system/src
[Coverage] -> /home/jenkins/.some-agent/workspace/system/src/main/java
[Coverage] -> finished painting successfully
[Coverage] -> zipping sources from folder '/home/jenkins/.some-agent/workspace/system/coverage' as '/home/jenkins/.some-agent/workspace/system/coverage-sources.zip'
[Coverage] Copying painted sources from agent to build folder
[Coverage] -> extracting...
[Coverage] -> done
[Coverage] Finished coverage processing - adding the action to the build...
```

3. Have a look at the temporary directory (`/tmp` in my case).

### Expected Results

Temporary data is being cleaned automatically, leaving no `coverage\d+` directories with each directory containing a rendered coverage HTML file behind.

### Actual Results

More than 2000 (randomly named) directories for each run which includes coverage reporting, which do not appear to be used at all. For Jenkins servers with tons of pipelines which are executed at a regular basis, this will quickly fill up the available disk space unnecessarily.

stefan6419846:
After the initial report, I did some research on how the painting actually is implemented and how the directories and files are used. It seems like the current implementation basically does the following for each source file:

* Create a temporary directory with a random name.
* Paint to a file with the transformed source file path inside this random directory.
* Compress the rendered file into a ZIP file with the same name inside the *coverage* sub-directory of the workspace.

[...]

Issue #433, opened 2022-07-22
IllegalStateException: Duplicate Key Error After Updating to 3.0.4


scothren:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
Jenkins: 2.346.2
OS: Linux - 5.13.0-1031-aws
---
ace-editor:1.1
ant:475.vf34069fef73c
antisamy-markup-formatter:2.7
apache-httpcomponents-client-4-api:4.5.13-1.0
authentication-tokens:1.4
aws-credentials:191.vcb_f183ce58b_9
aws-java-sdk-ec2:1.12.246-349.v96b_b_f7eb_a_c3c
aws-java-sdk-minimal:1.12.246-349.v96b_b_f7eb_a_c3c
bootstrap4-api:4.6.0-5
bootstrap5-api:5.1.3-7
bouncycastle-api:2.26
branch-api:2.1046.v0ca_37783ecc5
build-timeout:1.21
built-on-column:1.1
caffeine-api:2.9.3-65.v6a_47d0f4d1fe
checks-api:1.7.4
cloudbees-folder:6.729.v2b_9d1a_74d673
code-coverage-api:2.0.4
command-launcher:84.v4a_97f2027398
conditional-buildstep:1.4.2
config-file-provider:3.8.2
credentials:1139.veb_9579fca_33b_
credentials-binding:1.27.1
data-tables-api:1.12.1-2
dependencyanalyzer:0.7
disk-usage:0.28
display-url-api:2.3.6
docker-commons:1.19
docker-workflow:1.28
durable-task:496.va67c6f9eefa7
ec2:1.68
echarts-api:5.3.3-1
email-ext:2.87
envinject:2.866.v5c0403e3d4df
envinject-api:1.199.v3ce31253ed13
extended-choice-parameter:346.vd87693c5a_86c
extended-read-permission:3.2
external-monitor-job:191.v363d0d1efdf8
font-awesome-api:6.1.1-1
forensics-api:1.15.1
git:4.11.3
git-client:3.11.0
git-forensics:1.8.0
git-parameter:0.9.14
git-server:1.11
github:1.34.3
github-api:1.303-400.v35c2d8258028
github-branch-source:2.11.4
global-slack-notifier:1.5
gradle:1.39.4
handlebars:3.0.8
jackson2-api:2.13.3-285.vc03c0256d517
jacoco:3.3.2
javadoc:217.v905b_86277a_2a_
[...]

PR #442, opened 2022-08-04
Fix for Issue #433, #441 and URL anchors

fo-code:
This PR has been splitted:
see #443, #444, #445


plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageNode.java
```diff
@@ -255,6 +255,27 @@ public void add(final CoverageLeaf leaf) {
...
+     * Clears an empty tree path from the bottom of the tree to the top, beginning with the passed node.
+     *
+     * @param node The {@link CoverageNode node} to begin from
+     */
+    private void clearEmptyPaths(final CoverageNode node) {
+        if (node != null && node.getChildren().isEmpty()) {
+            node.remove();
+        }
+    }
+
```
uhafner, line 278: Is there a reason why you add wrong files and remove them afterwards? Wouldn't it make more sense to handle this while constructing the tree? Then we do not make the model mutable. This is something I would like to avoid.

uhafner (commented):
Can you please split this PR in *one* PR for *one* fix? It makes it hardly impossible to review the code.

fo-code:
@uhafner I see, I will do that in the future - here, I uploaded the whole coding session at once.

For this one, I added a description which files belong to which issue, these few files are clearly seperated (see above).
Would that be enough just this once?

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageNode.java
fo-code: Well yes, I totally agree with you.
In my opinion, the whole logic should be unnecessary and be removed after the problem with the ambiguous paths has been fixed.
I chose this option since it also relates to the reference trees, which belongs to older builds, where duplicate paths are also possible at the moment.
And even if I would simply not add files with duplicate paths, there is a possibility that the coverage tree contains empty packages afterwards without file nodes - in my opinion, there shouldn't be empty package nodes.

PR #444, opened 2022-08-07
[#441] Delete temporary folders on agent after coloring source code files

fo-code:
Fix for issue #441:

Folders and files which are temporary created while painting the source code are removed after usage now.
I also added an additional assertion to verify that the created temporary directories and only these are deleted.

### DIFF
--- a/plugin/src/test/java/io/jenkins/plugins/coverage/model/CoveragePluginSourceITest.java
+++ b/plugin/src/test/java/io/jenkins/plugins/coverage/model/CoveragePluginSourceITest.java
@@ -4,7 +4,6 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.util.Collections;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


plugin/src/test/java/io/jenkins/plugins/coverage/model/CoveragePluginSourceITest.java
```diff
@@ -115,14 +117,16 @@ private String createExternalSourceFolder() throws IOException {
         WorkflowJob job = createPipelineWithWorkspaceFiles(ACU_COBOL_PARSER_COVERAGE_REPORT);
         copyFileToWorkspace(job, SOURCE_FILE, checkoutDirectory + PACKAGE_PATH + "AcuCobolParser.java");
 
+        File temporaryDirectory = Paths.get("target", "tmp").toFile();
+        assertThat(temporaryDirectory.exists()).isTrue();
+        assertThat(temporaryDirectory.isDirectory()).isTrue();
+        File[] temporaryFiles = temporaryDirectory.listFiles();
+
```
uhafner, line 124: What is this directory? Does a test case store results there?

fo-code: This is the directory which is used as temporary directory /tmp within unit tests.
I check this directory in order to verify that painting source code does not store unused files there.
In this case, `Files.createTempDirectory(...)` uses this.

Nonetheless, I changed this line of code to make it more clear.

uhafner: Hmm, did you check that the test fails before you applied the fix (the RED during TDD?)? From my understanding, the directories will be part of the Jenkins agent file structure that does use a different directory. 

(I think it would be sufficient that the logging statement ist present)

fo-code: Yes I did. I made the tests fail, then I inserted the new code and it worked. Also, I debugged it and even checked the directories while doing this to make sure these temporary folders are created exacly there.

Also as I understand it, `Files.createTempDirectory(...)` points to `System.getProperty("java.io.tmpdir")` by default, if it is not overwritten using a unit test rule e.g. - also see [here](https://www.baeldung.com/java-temp-directories#delegate-to-the-operating-system).

uhafner: Yes, you are right. All tests use the controller to execute the builds. (Only the docker based tests would use a temporary folder in the docker container).

uhafner (approved):

PR #445, opened 2022-08-07
[#433] Remove duplicate file nodes 

fo-code:
Fixes issue #433:

I inserted a validation of the file structure.
`FileCoverageNodes` which have the same package and file name are removed from the coverage report since at the moment, these files cannot be handled correctly.
These cases will be logged including the name of the affected files.
This can be used as a workaround for the moment to prevent exceptions.
To also cover reference builds, this is executed for the current and the reference coverage tree.

I chose the option to delete ambiguous nodes afterwards, since it also relates to the reference trees, which belongs to older builds, where duplicate paths are also possible at the moment.
And even if I would simply not add files with duplicate paths when creating the tree, there is a possibility that the coverage tree contains empty packages afterwards without file nodes - in my opinion, there shouldn't be empty package nodes.
These methods which remove nodes should be removed when paths are handled as expected.

### DIFF
--- a/plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageNode.java
+++ b/plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageNode.java
@@ -256,6 +256,29 @@ public class CoverageNode implements Serializable {
     }
 
     /**
+     * Removes this node from the coverage tree.
+     */
+    void remove() {
+        // TODO: remove this method when unique paths are handled correctly
+        if (hasParent()) {
+            getParent().getChildren().remove(this);
+            clearEmptyPaths(getParent());
+        }
+    }
+
+    /**
+     * Clears an empty tree path from the bottom of the tree to the top, beginning with the passed node.
+     *
+     * @param node The {@link CoverageNode node} to begin from
+     */
+    private void clearEmptyPaths(final CoverageNode node) {
+        // TODO: remove this method when unique paths are handled correctly
+        if (node != null && node.getChildren().isEmpty()) {
+            node.remove();
+        }
+    }
+
+    /**
      * Returns whether this node is the root of the tree.
      *
      * @return {@code true} if this node is the root of the tree, {@code false} otherwise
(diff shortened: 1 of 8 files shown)


plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageNode.java
```diff
@@ -255,6 +255,29 @@ public void add(final CoverageLeaf leaf) {
         leaves.add(leaf);
     }
 
+    /**
+     * Removes this node from the coverage tree.
+     */
+    void remove() {
+        // TODO: remove this method when unique paths are handled correctly
+        if (hasParent()) {
+            getParent().getChildren().remove(this);
```
uhafner, line 264: `getChildren` is currently broken and will be readonly in a future version of the API.
```suggestion
            getParent().children.remove(this);
```

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageNode.java
```diff
@@ -255,6 +255,29 @@ public void add(final CoverageLeaf leaf) {
...
+    /**
+     * Clears an empty tree path from the bottom of the tree to the top, beginning with the passed node.
+     *
+     * @param node The {@link CoverageNode node} to begin from
+     */
+    private void clearEmptyPaths(final CoverageNode node) {
+        // TODO: remove this method when unique paths are handled correctly
+        if (node != null && node.getChildren().isEmpty()) {
+            node.remove();
+        }
```
uhafner, line 278: How can this happen?

This will also remove nodes that have a leaf.

uhafner (commented):
Is it really possible that we have an illegal structure in existing builds? I thought in this case the exception of #433 has been thrown? (Then we can skip all duplicates in `CoverageNodeConverter#createNode`)

(I wonder if it is better to )

uhafner (approved):
Since we need to adapt the API of the model anyway when we integrate the new parsers it does not hurt if we change the API right now.

uhafner:
I tested all changes locally and it looks fine! Thank you very much for your help!

Issue #449, opened 2022-08-15
Line coverage map is poorly readable


argorain:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
Jenkins: 2.346.3
OS: Windows Server 2016 - 10.0
---
Office-365-Connector:4.17.0
ace-editor:1.1
active-directory:2.25.1
ant:475.vf34069fef73c
antisamy-markup-formatter:2.7
apache-httpcomponents-client-4-api:4.5.13-138.v4e7d9a_7b_a_e61
authentication-tokens:1.4
authorize-project:1.4.0
azure-container-agents:241.va_780fa_dc374a_
azure-credentials:216.ve0b_4a_485ffc2
azure-sdk:118.v43f74dd9ca_dc
bitbucket:223.vd12f2bca5430
bitbucket-oauth:0.12
blueocean:1.25.6
blueocean-autofavorite:1.2.5
blueocean-bitbucket-pipeline:1.25.6
blueocean-commons:1.25.6
blueocean-config:1.25.6
blueocean-core-js:1.25.6
blueocean-dashboard:1.25.6
blueocean-display-url:2.4.1
blueocean-events:1.25.6
blueocean-git-pipeline:1.25.6
blueocean-github-pipeline:1.25.6
blueocean-i18n:1.25.6
blueocean-jwt:1.25.6
blueocean-personalization:1.25.6
blueocean-pipeline-api-impl:1.25.6
blueocean-pipeline-editor:1.25.6
blueocean-pipeline-scm-api:1.25.6
blueocean-rest:1.25.6
blueocean-rest-impl:1.25.6
blueocean-web:1.25.6
bootstrap4-api:4.6.0-5
bootstrap5-api:5.2.0-1
bouncycastle-api:2.26
branch-api:2.1046.v0ca_37783ecc5
build-timeout:1.21
caffeine-api:2.9.3-65.v6a_47d0f4d1fe
checks-api:1.7.5
cloud-stats:0.27
cloudbees-bitbucket-branch-source:785.ve724eb_44e286
cloudbees-folder:6.740.ve4f4ffa_dea_54
code-coverage-api:3.0.5
command-launcher:84.v4a_97f2027398
copyartifact:1.46.4
cppcheck:1.26
credentials:1139.veb_9579fca_33b_
credentials-binding:523.vd859a_4b_122e6
dark-theme:185.v276b_5a_8966a_e
[...]

PR #455, opened 2022-08-21
Fix for #449 - Enhanced tree map visualization

fo-code:
In order to make the tree map easier to read, I changed the following:

- the tree map is shown on the whole screen now 
- the border color saturation is alternating, following the [echarts documentation](https://echarts.apache.org/en/option.html#series-treemap.itemStyle.borderColorSaturation), which says that an alternating high color saturation is a good option for distinguishing different sections.

### DIFF
--- a/plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageViewModel/index.jelly
+++ b/plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageViewModel/index.jelly
@@ -3,7 +3,7 @@
 
   <st:header name="Content-Type" value="text/html;charset=UTF-8"/>
 
-  <bs:page it="${it}" class="fluid-container d-flex flex-column">
+  <bs:page it="${it}" class="fluid-container d-flex flex-column h-100">
 
     <st:adjunct includes="io.jenkins.plugins.echarts"/>
     <st:adjunct includes="io.jenkins.plugins.data-tables-select"/>
@@ -42,35 +42,34 @@
       </j:if>
     </ul>
 
-    <div class="tab-content">
+    <div class="tab-content h-100">
 
-      <div role="tabpanel" id="overview" class="tab-pane fade">
+      <div role="tabpanel" id="overview" class="tab-pane fade h-100">
+
         <div class="row">
-
           <div class="col-5">
             <bs:card title="${%Total coverage overview}" fontAwesomeIcon="chart-bar">
-              <div id="coverage-overview" class="graph-cursor-pointer overview-chart h-100"/>
+              <div id="coverage-overview" class="graph-cursor-pointer overview-chart"/>
             </bs:card>
           </div>
-
           <div class="col-7">
             <bs:card title="${%Coverage trend}" fontAwesomeIcon="chart-line">
-              <div id="coverage-trend" class="graph-cursor-pointer overview-chart h-100"/>
+              <div id="coverage-trend" class="graph-cursor-pointer overview-chart"/>
             </bs:card>
           </div>
-
         </div>
       </div>
 
-      <div role="tabpanel" id="lineCoverage" class="tab-pane fade">
-        <bs:card title="${%Line coverage}" fontAwesomeIcon="folder-tree">
-          <div id="project-line-coverage" class="graph-cursor-pointer tree-chart"/>
+      <div role="tabpanel" id="lineCoverage" class="tab-pane fade h-100">
+        <bs:card title="${%Line coverage}" fontAwesomeIcon="folder-tree"
+                 class="flex-fill h-100" bodyClass="d-flex flex-column h-100">
+          <div id="project-line-coverage" class="graph-cursor-pointer tree-chart flex-grow-1"/>
         </bs:card>
       </div>
-
-      <div role="tabpanel" id="branchCoverage" class="tab-pane fade">
-        <bs:card title="${%Branch coverage}" fontAwesomeIcon="folder-tree" >
-          <div id="project-branch-coverage" class="graph-cursor-pointer tree-chart"/>
+      <div role="tabpanel" id="branchCoverage" class="tab-pane fade h-100">
+        <bs:card title="${%Branch coverage}" fontAwesomeIcon="folder-tree"
+                 class="flex-fill h-100" bodyClass="d-flex flex-column h-100">
+          <div id="project-branch-coverage" class="graph-cursor-pointer tree-chart flex-grow-1"/>
         </bs:card>
       </div>
 
--- a/plugin/src/main/webapp/js/charts.js
+++ b/plugin/src/main/webapp/js/charts.js
@@ -168,63 +168,63 @@ const CoverageChartGenerator = function ($) {
                     itemStyle: {
                         borderColor: '#ddd',
                         borderWidth: 2,
-                        gapWidth: 2
+                        gapWidth: 3
                     }
                 },
                 {
                     itemStyle: {
                         borderWidth: 4,
-                        gapWidth: 2,
-                        borderColorSaturation: 0.6
+                        gapWidth: 5,
+                        borderColorSaturation: 0.2
                     }
                 },
                 {
                     itemStyle: {
                         borderWidth: 4,
-                        gapWidth: 2,
-                        borderColorSaturation: 0.7
+                        gapWidth: 3,
+                        borderColorSaturation: 0.8
                     }
                 },
                 {
                     itemStyle: {
                         borderWidth: 4,
-                        gapWidth: 2,
-                        borderColorSaturation: 0.6
+                        gapWidth: 3,
+                        borderColorSaturation: 0.2
                     }
                 },
                 {
                     itemStyle: {
                         borderWidth: 4,
-                        gapWidth: 2,
-                        borderColorSaturation: 0.7
+                        gapWidth: 3,
+                        borderColorSaturation: 0.8
                     }
                 },
                 {
                     itemStyle: {
                         borderWidth: 4,
-                        gapWidth: 2,
-                        borderColorSaturation: 0.6
+                        gapWidth: 3,
+                        borderColorSaturation: 0.2
                     }
                 },
                 {
                     itemStyle: {
                         borderWidth: 4,
-                        gapWidth: 2,
-                        borderColorSaturation: 0.7
+                        gapWidth: 3,
+                        borderColorSaturation: 0.8
                     }
                 },
                 {
                     itemStyle: {
                         borderWidth: 4,
-                        gapWidth: 2,
-                        borderColorSaturation: 0.6
+                        gapWidth: 3,
+                        borderColorSaturation: 0.2
                     }
                 },
                 {
                     itemStyle: {
                         borderWidth: 4,
-                        gapWidth: 2,
-                        borderColorSaturation: 0.7
+                        gapWidth: 3,
+                        borderColorSaturation: 0.8
                     }
                 },
             ];


plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageViewModel/index.jelly
```diff
@@ -3,7 +3,7 @@
 
   <st:header name="Content-Type" value="text/html;charset=UTF-8"/>
 
-  <bs:page it="${it}" class="fluid-container d-flex flex-column">
+  <bs:page it="${it}" class="fluid-container d-flex flex-column h-100">
```
uhafner, line 6: This breaks the visualization on smaller displays.

plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageViewModel/index.jelly
```diff
@@ -42,35 +42,34 @@
       </j:if>
     </ul>
 
-    <div class="tab-content">
+    <div class="tab-content h-100">
```
uhafner, line 45: See above

plugin/src/main/webapp/js/charts.js
```diff
@@ -168,63 +168,63 @@ const CoverageChartGenerator = function ($) {
...
+                        borderColorSaturation: 0.2
                     }
                 },
                 {
                     itemStyle: {
                         borderWidth: 4,
-                        gapWidth: 2,
-                        borderColorSaturation: 0.7
+                        gapWidth: 3,
+                        borderColorSaturation: 0.8
```
uhafner, line 227: In total I think the greens look better, but the reds not.

uhafner (approved):
Would it help to use white as text color?

uhafner:
> the tree map is shown on the whole screen now
The problem with that approach is, that you then need to scroll on small displays. I found no way yet to always use the whole visible screen only for the trees. Maybe the problem is that the side panel is too large.

uhafner (changes requested):
> the tree map is shown on the whole screen now

The problem with that approach is, that you then need to scroll on small displays. I found no way yet to always use the whole visible screen only for the trees. Maybe the problem is that the side panel is too large.

fo-code:
> The problem with that approach is, that you then need to scroll on small displays. I found no way yet to always use the whole visible screen only for the trees. Maybe the problem is that the side panel is too large.

Okay, well I do not have a small screen and when I try it by resizing the window, it fills just the visible screen due to the resize event, but this might behave differently on a real small screen.
Nonetheless, I think in most cases `h-100` is useful, even if a few users have to scroll. 
I guess not that many developer use a small laptop screen all the time and even if it is the case, a bigger tree map allows that more file names are displays and the scrolling is limited.
What do you think about that?

> In total I think the greens look better, but the reds not.

I totally agree - I tried some different approaches and found a better solution. 
Also, I recognized that the package colors are generated by the echarts API, but these should be set dependend on the coverage.
Furthermore, the alpha of the currently used colors are problematic in the case of the map, so I removed these.
Therefore, I locally changed some values including the echarts plugin and the result can be shown above - I updated the "after" examples.
The new approach shows the coverage of each node exactly and different levels are clearly separable.

I am not available for two weeks now, so I would update the code afterwards - but maybe you can already give me some feedback on the new approach?

uhafner:
> > The problem with that approach is, that you then need to scroll on small displays. I found no way yet to always use the whole visible screen only for the trees. Maybe the problem is that the side panel is too large.
> 
> Okay, well I do not have a small screen and when I try it by resizing the window, it fills just the visible screen due to the resize event, but this might behave differently on a real small screen. 
> Nonetheless, I think in most cases `h-100` is useful, even if a few users have to scroll. I guess not that many developer use a small laptop screen all the time and even if it is the case, a bigger tree map allows that more file names are displays and the scrolling is limited. What do you think about that?

I would rather prefer to create different views for small and large screens. This shouldn't be too hard with Bootstraps classes.  See https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/plugin/src/main/resources/coverage/coverage-table.jelly#L14 for an example

(See https://getbootstrap.com/docs/5.2/layout/grid/ for a documentation of all responsive classes)

> 
> > In total I think the greens look better, but the reds not.
> 
> I totally agree - I tried some different approaches and found a better solution. Also, I recognized that the package colors are generated by the echarts API, but these should be set dependend on the coverage. Furthermore, the alpha of the currently used colors are problematic in the case of the map, so I removed these. Therefore, I locally changed some values including the echarts plugin and the result can be shown above - I updated the "after" examples. The new approach shows the coverage of each node exactly and different levels are clearly separable.
> 
> I am not available for two weeks now, so I would update the code afterwards - but maybe you can already give me some feedback on the new approach?

The new colors look better now. I wonder if we should use the red, orange and green colors of Jenkins as well here? 
https://weekly.ci.jenkins.io/design-library/Colors/

fo-code:
> I would rather prefer to create different views for small and large screens.

So you mean we can add `h-100` for large screens only in order to use the full screen here but also support small screens without `h-100`?


> I wonder if we should use the red, orange and green colors of Jenkins as well here?

Yes of course, so you mean to change the currently used colors in `CoverageColorPalette` and insert the RGB codes of the Jenkins colors manually? Or is there a possibility to load the library colors dynamically as for example RGB values in Java?

uhafner:
> > I would rather prefer to create different views for small and large screens.
> 
> So you mean we can add `h-100` for large screens only in order to use the full screen here but also support small screens without `h-100`?
 
Yes, that shouldn't be too hard to implement.

> > I wonder if we should use the red, orange and green colors of Jenkins as well here?
> 
> Yes of course, so you mean to change the currently used colors in `CoverageColorPalette` and insert the RGB codes of the Jenkins colors manually? Or is there a possibility to load the library colors dynamically as for example RGB values in Java?

Getting them dynamically should work for the charts if we change the proxy methods:

So rather than using
```
            viewProxy.getOverview(function (t) {
                createOverview(t.responseObject(), 'coverage-overview');
            });
```
we could use

```
            viewProxy.getOverview(colors, function (t) {
                createOverview(t.responseObject(), 'coverage-overview');
            });
```

and fill `colors` with a JSON object, that contains the important styles. Something like:
```
{
 green: '#00ff00',
 orange: '#00ff00',
 red: '#00ff00'
}
```
(See https://weekly.ci.jenkins.io/design-library/Colors/ for the available colors).

You can get the colors then via:
```
    function getTextColor() {
        return getComputedStyle(document.body).getPropertyValue('--text-color') || '#333';
    }
```

Then the colors automatically will be adapted to the selected theme.

fo-code:
> Yes, that shouldn't be too hard to implement.

I agree, this looks like a good solution.


> Getting them dynamically should work for the charts if we change the proxy methods:

Thanks for the explanation.
As I understand this, the color JSON has to be set manually? In this case, I prefer using an Enum since this allows using these colors within the whole plugin and it is easier to integrate.
More specific what I wanted to know is if for example this color JSON can be requested or if I can call a Java API to request these color hex codes within the Java code of the coverage plugin? I am not really familiar of how the design library works...

uhafner:
> > Yes, that shouldn't be too hard to implement.
> 
> I agree, this looks like a good solution.
> 
> > Getting them dynamically should work for the charts if we change the proxy methods:
> 
> Thanks for the explanation. As I understand this, the color JSON has to be set manually? In this case, I prefer using an Enum since this allows using these colors within the whole plugin and it is easier to integrate. More specific what I wanted to know is if for example this color JSON can be requested or if I can call a Java API to request these color hex codes within the Java code of the coverage plugin? I am not really familiar of how the design library works...

It works the other was round: there is no way for the server to call the client (and there is no API on the client side). You need to wrap all information in an existing Ajax call from the client to the server:

In Jenkins processing is started on the client where the user selects the HTTP address. This is mapped on the server side into a jelly view. This jelly view is then rendered as HTML and send to the client that will render it. In this view we can invoke an associated JS file and call back to the server using Ajax. I.e., when the client currently starts a request to get the coverage data (the chart is empty in the jelly view) the client can provide an additional parameter (that is the reason that I am proposing a JSON object on the JS side, where we can put everything into what we want. I use the same thing in the trend chart configuration). This JSON can be interpreted on the server side to extract the colors. An enum will not work as it is constant on the server side, but actually each client might have its own colors. Maybe we can setup a call to discuss this part in more detail?

fo-code:
Yes, this would be perfect! I sent you an email.

fo-code:
@uhafner I finished the color rework.
The Jenkins colors from the Design Library are now in use and for the table and the tree map dynamically loaded.
The fallback colors correspond to the Jenkins colors aswell.
Currently, some HSL colors are provided in the dark mode - these cases are not handled yet and the fallback is used.
The size of the tree map is like before in order to work on all screen sizes.

plugin/src/main/java/io/jenkins/plugins/coverage/model/CoverageTableModel.java
```diff
@@ -206,12 +215,12 @@ public int getLoc() {
...
             if (coverage.isSet()) {
                 double percentage = coverage.getCoveredPercentage().getDoubleValue();
-                DisplayColors colors = CoverageLevel.getDisplayColorsOfCoverageLevel(percentage, COLOR_PROVIDER);
+                DisplayColors colors = CoverageLevel.getDisplayColorsOfCoverageLevel(percentage, colorProvider);
                 String cell = div().withClasses(COVERAGE_COLUMN_OUTER).with(
                                 div().withClasses(COVERAGE_COLUMN_INNER)
                                         .withStyle(String.format(
                                                 "background-image: linear-gradient(90deg, %s %f%%, transparent %f%%);",
-                                                colors.getFillColorAsHex(),
+                                                colors.getFillColorAsRGBAHex(80),
```
uhafner, line 223: 80 seems to be some kind of random, maybe it helps to define a constant with this value that uses a appropriate name?

plugin/src/main/java/io/jenkins/plugins/coverage/model/visualization/colorization/CoverageColorJenkinsId.java
```diff
@@ -0,0 +1,45 @@
...
+    LIGHT_RED("--light-red");
+
+    private final String jenkinsColorId;
+
+    /**
+     * Constructor.
+     *
+     * @param colorId
+     *         The color id
+     */
```
uhafner, line 31: This can be deleted

plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageViewModel/index.jelly
uhafner: Can we remove these unrelated changes? Then we can merge the color PR and create a new release.

plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageViewModel/index.jelly
```diff
@@ -44,35 +45,34 @@
       </ul>
     </div>
 
-    <div class="tab-content">
+    <div class="tab-content pt-2">
```
uhafner, line 48: See above

plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageViewModel/index.jelly
```diff
@@ -44,35 +45,34 @@
...
 
-      <div role="tabpanel" id="overview" class="tab-pane fade" aria-labelledby="overview-tab" >
+      <div role="tabpanel" id="overview" class="tab-pane fade h-100" aria-labelledby="overview-tab" >
```
uhafner, line 50: See above

plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageViewModel/index.jelly
```diff
@@ -44,35 +45,34 @@
...
         <div class="row">
-
           <div class="col-5">
             <bs:card title="${%Total coverage overview}" fontAwesomeIcon="chart-bar">
-              <div id="coverage-overview" class="graph-cursor-pointer overview-chart h-100"/>
+              <div id="coverage-overview" class="graph-cursor-pointer overview-chart"/>
```
uhafner, line 54: See above

plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageViewModel/index.jelly
```diff
@@ -44,35 +45,34 @@
...
             </bs:card>
           </div>
-
           <div class="col-7">
             <bs:card title="${%Coverage trend}" fontAwesomeIcon="chart-line">
-              <div id="coverage-trend" class="graph-cursor-pointer overview-chart h-100"/>
+              <div id="coverage-trend" class="graph-cursor-pointer overview-chart"/>
```
uhafner, line 59: See above

plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageViewModel/index.jelly
```diff
@@ -44,35 +45,34 @@
...
         </div>
       </div>
 
-      <div role="tabpanel" id="lineCoverage" class="tab-pane fade" aria-labelledby="lineCoverage-tab">
-        <bs:card title="${%Line coverage}" fontAwesomeIcon="folder-tree">
-          <div id="project-line-coverage" class="graph-cursor-pointer tree-chart"/>
+      <div role="tabpanel" id="lineCoverage" class="tab-pane fade h-100" aria-labelledby="lineCoverage-tab">
+        <bs:card title="${%Line coverage}" fontAwesomeIcon="folder-tree"
+                 class="flex-fill h-100" bodyClass="d-flex flex-column h-100">
+          <div id="project-line-coverage" class="graph-cursor-pointer tree-chart flex-grow-1"/>
```
uhafner, line 68: See above

plugin/src/main/resources/io/jenkins/plugins/coverage/model/CoverageViewModel/index.jelly
```diff
@@ -44,35 +45,34 @@
...
         </bs:card>
       </div>
 
-      <div role="tabpanel" id="branchCoverage" class="tab-pane fade" aria-labelledby="branchCoverage-tab">
-        <bs:card title="${%Branch coverage}" fontAwesomeIcon="folder-tree">
-          <div id="project-branch-coverage" class="graph-cursor-pointer tree-chart"/>
+      <div role="tabpanel" id="branchCoverage" class="tab-pane fade h-100" aria-labelledby="branchCoverage-tab">
+        <bs:card title="${%Branch coverage}" fontAwesomeIcon="folder-tree"
+                 class="flex-fill h-100" bodyClass="d-flex flex-column h-100">
+          <div id="project-branch-coverage" class="graph-cursor-pointer tree-chart flex-grow-1"/>
```
uhafner, line 75: See above

uhafner (changes requested):
I like the new design! The only mandatory changes are the viewpoint adjustments which are unrelated to this PR and should be changed in a different PR.

fo-code:
Nice to hear and thanks for the review!
Then, everything should be ready to be released. :)

uhafner (approved):

PR #486, opened 2022-09-26
Prevent mismatches between code delta and jacoco data - Issue #467

fo-code:
I added an additional check to prevent exception in case of mismatches between the calculated code delta and the Jacoco coverage data.
This prevents the bug from issue #467.
After an unexpected behavior has been detected, all caluclations based on the code delta are skipped and the error is logged.

### DIFF
--- a/plugin/src/main/java/io/jenkins/plugins/coverage/model/CodeDeltaCalculator.java
+++ b/plugin/src/main/java/io/jenkins/plugins/coverage/model/CodeDeltaCalculator.java
@@ -14,6 +14,8 @@ import java.util.stream.Collectors;
 
 import edu.hm.hafner.util.FilteredLog;
 
+import one.util.streamex.StreamEx;
+
 import hudson.FilePath;
 import hudson.model.Run;
 import hudson.model.TaskListener;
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/io/jenkins/plugins/coverage/model/CodeDeltaCalculator.java
```diff
@@ -229,4 +236,41 @@ private void verifyScmToReportPathMapping(final Map<String, String> pathMapping,
...
+                .distinct(2)
+                .collect(Collectors.toSet());
+        Map<String, String> duplicateEntries = new HashMap<>();
+        for (Map.Entry<String, String> entry : oldPathMapping.entrySet()) {
+            for (String duplicate : duplicates) {
+                if (entry.getValue().equals(duplicate)) {
+                    duplicateEntries.put(entry.getKey(), duplicate);
+                }
+            }
+        }
```
uhafner, line 265: Wouldn't it make more sense to move that part of the code into the `if` block?

plugin/src/main/java/io/jenkins/plugins/coverage/model/CodeDeltaCalculator.java
```diff
@@ -229,4 +236,41 @@ private void verifyScmToReportPathMapping(final Map<String, String> pathMapping,
...
+        Set<String> duplicates = StreamEx.of(oldPathMapping.values())
+                .distinct(2)
+                .collect(Collectors.toSet());
+        Map<String, String> duplicateEntries = new HashMap<>();
+        for (Map.Entry<String, String> entry : oldPathMapping.entrySet()) {
+            for (String duplicate : duplicates) {
+                if (entry.getValue().equals(duplicate)) {
+                    duplicateEntries.put(entry.getKey(), duplicate);
+                }
+            }
```
uhafner, line 264: Simplify:
```
if (duplicates.contains(entry.getValue()) {
    duplicateEntries.put(entry.getKey(), entry.getValue());
}
```

(or even more functional)

uhafner (approved):
Since this is only debugging code, only some optional improvements.

Issue #496, opened 2022-10-04
Extend documentation with new features


fo-code:
The README does not contain any information about the new features since version 3.0.0.
It would be helpful to see the different use cases and how the plugin can be used.

PR #501, opened 2022-10-07
Feature documentation - Issue #496

fo-code:
I added a documentation of the new features since version 3.0.0 into the README, including example screenshots.
See issue #496 .


README.md
```diff
@@ -34,7 +29,45 @@ It has been developed during [GSoC 2018](https://jenkins.io/projects/gsoc/2018/c
...
-## How to Use it
+## Features
+### General Features
+* Pipeline support
+* Parallel execution in pipeline support
+* Reports combining
+* REST API
+* Failed conditions and flexible threshold setting
+
+### Coverage Analysis
```
uhafner, line 40: I think we can delete this part as well. Currently only pipeline support is working (and thresholds need to be changed as well).

uhafner (approved):

README.md
fo-code: Yes, you are right - done. 
Do you already know when there will be the first version of the new parsers? ... since a rework of the thresholds would be really helpful as well. :)

uhafner: > Do you already know when there will be the first version of the new parsers? ... since a rework of the thresholds would be really helpful as well. :)

The new parsers have been merged already in https://github.com/uhafner/coverage-model.

Ihm currently integrating them in 
https://github.com/jenkinsci/code-coverage-api-plugin/pull/502

Let's see how far I get the implementation today. Then I will make a release of `coverage-model`.

PR #502, opened 2022-10-08
Create a new step to record coverage results

uhafner:
Rewrite of the recorder (and step) so that we can use the new [Coverage Model](https://github.com/uhafner/coverage-model/blob/main/pom.xml). The old step is too tightly coupled with the old model and cannot be refactored so that old pipelines will remain compatible.

Downstream of https://github.com/uhafner/coverage-model/pull/35


fo-code:
Hi, I just wanted to ask about the status of this pull request. Is there anything I can do to help you with this? :)
I am just wondering because I thought about implementing thresholds for Delta Coverage and an API for accessing these information from other Plugins, but this depends on these model changes here.

uhafner:
Superseded by #512.

uhafner:
> Hi, I just wanted to ask about the status of this pull request. Is there anything I can do to help you with this? :)
> I am just wondering because I thought about implementing thresholds for Delta Coverage and an API for accessing these information from other Plugins, but this depends on these model changes here.

I answered in an email, I hope I used the correct address...

Issue #191, opened 2021-04-01
failBuildIfCoverageDecreasedInChangeRequest should not fail if no code changed


fuzzybair:
### Version report

Jenkins and plugins versions report:

```
Jenkins: 2.263.1
OS: Linux - 4.15.0-136-generic
---
sonar:2.12
blueocean-core-js:1.24.3
git:4.4.4
cobertura:1.16
[Many others as this is a shared server]
```

- What Operating System are you using (both controller, and any agents involved in the problem)?

```
OS: Linux - 4.15.0-136-generic
```

### Reproduction steps

We have several items that are not code related in our repository (e.g. documentation, images, etc.). when we have a pull request that is strictly non code the Pull Request can fail. The failure seems to be related to a rounding difference between the stored and calculated values.
Possible solutions:
* Add a level of tolerance to failBuildIfCoverageDecreasedInChangeRequest to account for the rounding.
* Always round down before storing the value
* Use the same precision with stored and calculated.

### Results

Here is our configuration for the code coverage.
```
      publishCoverage adapters: [coberturaAdapter(mergeToOneReport: false, path: 'codeCoverageReports/Cobertura.xml')],
        calculateDiffForChangeRequests: true, failNoReports: true, failUnhealthy: false, failUnstable: true, failBuildIfCoverageDecreasedInChangeRequest: true,
        globalThresholds: [[thresholdTarget: 'Line', unhealthyThreshold: 85.0, unstableThreshold: 80.0]], sourceFileResolver: sourceFiles('STORE_LAST_BUILD') 
```
Expected result:

failBuildIfCoverageDecreasedInChangeRequest should not fail if there are not any associated code changes. 

Actual result:

We have observed values indicated coverage changed on these pull request between -0.004% and +0.004% if the rounding rounds down from the calculated value the request will show an increase in coverage. But if the request follows after a merge that rounded up it will show a decrease in coverage, and therefore fail the check.

reshadatavid:
This is still a problem. Happening because of Floating point rounding error (happens in and other languages frequently).

uhafner:
> This is still a problem. Happening because of Floating point rounding error (happens in and other languages frequently).

I hope this will be fixed when the migration to the new model is fixed, see #512 for details.

reshadatavid:
Also is it possible to get coverage data in groovy. What I want to do is, instead of relying on fail via plugin, I would take up data in groovy and do necessary calculations

[...]

PR #533, opened 2022-12-16
Fix float precision issue in coverage diff 

reshadat:
While calculating difference in coverage, there might be a difference even without code change, due to floating point rounding errors. Converting the diff float to BigDecimal with appropriate scale and converting back solves this issue.

Fixes: https://github.com/jenkinsci/code-coverage-api-plugin/issues/191

### DIFF
--- a/plugin/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
+++ b/plugin/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
@@ -7,6 +7,8 @@ import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.PrintStream;
+import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.util.ArrayList;
@@ -252,7 +254,8 @@ public class CoverageProcessor {
     private void failBuildIfChangeRequestDecreasedCoverage(final CoverageResult coverageResult, final CoverageAction action)
             throws CoverageException {
         float coverageDiff = coverageResult.getCoverageDelta(CoverageElement.LINE);
-        if (coverageDiff < 0) {
+
+        if (roundFloat(coverageDiff,2) < 0) {
             String message = "Fail build because this change request decreases line coverage by " + coverageDiff;
             action.setFailMessage(message);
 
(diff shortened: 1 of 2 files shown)


plugin/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -227,7 +228,8 @@ private CoverageAction convertResultToAction(final CoverageResult coverageReport
             Ratio buildRatio = coverageReport.getCoverage(coverageElement);
 
             if (buildRatio != null) {
-                float diff = buildRatio.getPercentageFloat() - referenceRatio.getPercentageFloat();
+                //Converting float point to Big decimal and converting back fixes the float precision point problem
+                float diff = BigDecimal.valueOf((buildRatio.getPercentageFloat() - referenceRatio.getPercentageFloat())).floatValue();
```
uhafner, line 232: Can you wrap that code into a method? Then you can write a small unit test that verifies the new behavior.

uhafner (changes requested):

plugin/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
reshadat: Added tests. Please re-review

plugin/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -731,4 +732,13 @@ public static CoverageResult recoverCoverageResult(final Run<?, ?> run) throws I
             return (CoverageResult) ois.readObject();
         }
     }
+    /**
+     * Round up float value to specified scaling factor using Round down strategy
+     * @param number
+     * @param scale
+     * @return float value (scaled)
+     */
```
uhafner, line 740:
```suggestion
```

plugin/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
```diff
@@ -731,4 +732,13 @@ public static CoverageResult recoverCoverageResult(final Run<?, ?> run) throws I
...
+    public static float roundFloat(final float number, final int scale) {
```
uhafner, line 741:
```suggestion
    static float roundFloat(final float number, final int scale) {
```

uhafner (approved):

reshadatavid:
@uhafner  What should be the next steps? Should I do something else for this to be merged and releases in new verison?

uhafner:
No, I just. needed to wait for Jenkins to mark this PR as green. 

Since this code is deprecated anyway there is no need to fix styling issues (I already removed the corresponding checks).

PR #512, opened 2022-11-06
Redesign of plugin using external coverage model

uhafner:
Refactoring of the recorder and step using the external [Coverage Model](https://github.com/uhafner/coverage-model).

Since we cannot make the new API (moving of the model classes to an external component, refactoring of the steps parameters, etc.) backward compatible, a new recorder (and step) has been added. The source code has been copied to the new `metrics` package so that old pipelines will still work out of the box. In order to use the new features, a new recorder needs to be used. The old code will be marked as deprecated and will not be supported anymore. It might get removed in a future version of this plugin.

## Coverage of delta changes 

The plugin now computes and shows the absolute coverage of the project, the coverage of the modified files and the coverage of the modified lines so you can see how the changes actually affect the code coverage. Additionally, the delta of these coverages with respect to the reference build are computed.

Bildschirm­foto 2023-03-09 um 08 24 55
 
## Cyclomatic Complexity and LOC metrics

Several coverage parsers support the measurement of cyclomatic complexity and lines of code. These metrics are now computed and recorded as well:

Bildschirm­foto 2023-01-13 um 14 48 53

Bildschirm­foto 2023-01-20 um 23 27 03

## Support for mutation coverage (PIT)

The new model has been extended so that not only code coverage but also mutation coverage can be evaluated and stored. That means that you can visualize JaCoCo code coverage results along with PIT mutation testing results. So a single code base and pipeline configuration can be used for these different report formats:

Bildschirm­foto 2023-01-17 um 22 15 07

Bildschirm­foto 2023-01-20 um 23 26 43

Bildschirm­foto 2023-01-25 um 23 27 05

Bildschirm­foto 2023-02-05 um 17 39 27

## Display of all Metrics in tooltips

The coverage column and the build summary will show all available metrics in a tooltip. 

Bildschirm­foto 2023-02-02 um 07 55 22

## Quality Gates

You can now specify an arbitrary number of quality gates that allow to set the build to unstable or failed if the thresholds are not met. For each quality gate the metric (branch coverage, complexity, etc.) and the baseline (whole project, changed files, etc.) can be defined. 

Bildschirm­foto 2022-12-30 um 11 42 16

Bildschirm­foto 2023-01-08 um 23 11 47

## Integrate new model into Checks report

The plugin continues to publish checks to supporting SCM platforms. Now the details of the checks report have been updated to be in sync with the details of the Jenkins build results.

Bildschirm­foto 2023-03-08 um 12 22 58

[...]

PR #556, opened 2023-02-16
Add Modified Files Coverage and Checks Publisher

fo-code:
I added Modified Files Coverage support and a new Coverage Checks Publisher including tests for it.
Also, I renamed "Change Coverage" to "Modified Lines Coverage".

Example layout for the Checks Overview:
grafik


plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/ElementFormatter.java
```diff
@@ -370,6 +376,51 @@ public String getDisplayName(final Metric metric) {
...
 
+    /**
+     * Gets the display names of the existing {@link Metric coverage metrics}, sorted by the metrics ordinal.
+     *
+     * @return the sorted metric display names
+     */
+    public List<String> getSortedCoverageDisplayNames() {
+        return Metric.getCoverageMetrics().stream()
+                // use the default comparator which uses the enum ordinal
+                .sorted()
```
uhafner, line 387: Isn't the set already sorted in the correct way?

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/ElementFormatter.java
```diff
@@ -370,6 +376,51 @@ public String getDisplayName(final Metric metric) {
...
+                .map(this::getDisplayName)
+                .collect(Collectors.toList());
+    }
+
+    /**
+     * Formats a stream of values to theis display representation by using the passed locale.
```
uhafner, line 393:
```suggestion
     * Formats a stream of values to their display representation by using the given locale.
```

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/ElementFormatter.java
```diff
@@ -370,6 +376,51 @@ public String getDisplayName(final Metric metric) {
...
+     */
+    public Stream<Coverage> getSortedCoverageValues(final Node coverage) {
+        return Metric.getCoverageMetrics()
+                .stream()
+                .map(m -> m.getValueFor(coverage))
+                .flatMap(Optional::stream)
+                .filter(value -> value instanceof Coverage)
+                .map(Coverage.class::cast)
+                .sorted(Comparator.comparing(Coverage::getMetric));
+    }
```
uhafner, line 422: same here

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/source/SourceCodeFacade.java
```diff
@@ -163,7 +163,7 @@ File createFileInBuildFolder(final File buildResults, final String id, final Str
     }
 
     /**
-     * Filters the sourcecode coverage highlighting for analyzing the change coverage only.
+     * Filters the sourcecode coverage highlighting for analyzing the Modified Lines Coverage only.
```
uhafner, line 166:
```suggestion
     * Filters the sourcecode coverage highlighting for analyzing the modified lines coverage only.
```

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageBuildAction.java
```diff
@@ -343,6 +385,39 @@ public boolean hasDelta(final Baseline baseline, final Metric metric) {
...
+     * @param metric
+     *         the metric to get the delta for
+     *
+     * @return the formatted value
+     */
+    public String formatValue(final Baseline baseline, final Metric metric) {
+        var value = getAllValues(baseline).stream()
+                .filter(v -> v.getMetric() == metric)
+                .findFirst();
+        return value.isPresent() ? FORMATTER.formatValue(value.get()) : Messages.Coverage_Not_Available();
```
uhafner, line 418: This should be actually handled by the formatter

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageRecorder.java
```diff
@@ -63,6 +63,9 @@
  */
 @SuppressWarnings("checkstyle:ClassFanOutComplexity")
 public class CoverageRecorder extends Recorder {
+    // TODO: make customizable?
```
uhafner, line 66: Yes, I think someone else added a PR to make that configurable in the existing publisher.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageReporter.java
```diff
@@ -138,16 +148,16 @@ private void createDeltaReports(final Node rootNode, final FilteredLog log, fina
         catch (IllegalStateException exception) {
             log.logError("An error occurred while processing code and coverage changes:");
             log.logError("-> Message: " + exception.getMessage());
-            log.logError("-> Skipping calculating change coverage and indirect coverage changes");
+            log.logError("-> Skipping calculating Modified Lines Coverage and indirect coverage changes");
```
uhafner, line 151:
```suggestion
            log.logError("-> Skipping calculating modified lines coverage and indirect coverage changes");
```

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageTableModel.java
```diff
@@ -115,15 +115,15 @@ public List<TableColumn> getColumns() {
                 .withType(ColumnType.NUMBER)
                 .build();
         columns.add(loc);
-        if (root.containsMetric(Metric.COMPLEXITY)) {
+        if (root.getMetrics().contains(Metric.COMPLEXITY)) {
```
uhafner, line 118: Why that? The previous code is more object oriented

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageViewModel.java
```diff
@@ -407,11 +412,11 @@ public boolean hasSourceCode() {
     }
 
     /**
-     * Checks whether change coverage exists.
+     * Checks whether Modified Lines Coverage exists.
```
uhafner, line 415:
```suggestion
     * Checks whether modified lines coverage exists.
```

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageViewModel.java
```diff
@@ -407,11 +412,11 @@ public boolean hasSourceCode() {
...
      *
-     * @return {@code true} whether change coverage exists, else {@code false}
+     * @return {@code true} whether Modified Lines Coverage exists, else {@code false}
```
uhafner, line 417:
```suggestion
     * @return {@code true} whether modified lines coverage exists, else {@code false}
```

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/ModifiedFilesCoverageTable.java
```diff
@@ -0,0 +1,109 @@
...
+                .filter(node -> node.getPath().equals(fileNode.getPath())
+                        && node.getName().equals(fileNode.getName()))
+                .findFirst()
+                .orElse(fileNode); // return this as fallback to prevent exceptions
+    }
+
+    /**
+     * UI row model for the modified lines coverage details table.
+     *
+     * @since 4.0.0
```
uhafner, line 73:
```suggestion
```

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/ModifiedFilesCoverageTable.java
```diff
@@ -0,0 +1,109 @@
...
+
+import io.jenkins.plugins.coverage.metrics.color.ColorProvider;
+import io.jenkins.plugins.datatables.DetailedCell;
+import io.jenkins.plugins.datatables.TableConfiguration;
+import io.jenkins.plugins.datatables.TableConfiguration.SelectStyle;
+
+/**
+ * {@link CoverageTableModel} implementation for visualizing the modified lines coverage.
+ *
+ * @since 4.0.0
```
uhafner, line 22: Since is only used for API methods
```suggestion
```

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/ModifiedFilesCoverageTable.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+class ModifiedFilesCoverageTable extends CoverageTableModel {
```
uhafner, line 24: Shouldn't this class extend from `ChangesTableModel`?

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageViewModel.java
```diff
@@ -407,11 +412,11 @@ public boolean hasSourceCode() {
...
      */
-    public boolean hasChangeCoverage() {
+    public boolean hasModifiedLinesCoverage() {
```
uhafner, line 419: These changes prevent that the table is shown in the UI. I think a change to the `index.jelly` is missing.

uhafner (commented):
I will continue the review tomorrow with the checks publisher and the tests...

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/Baseline.java
```diff
@@ -27,23 +27,23 @@ public enum Baseline {
...
      * focus on new or modified code only.
      */
-    MODIFIED_FILES(Messages._Baseline_MODIFIED_FILES(), "fileCoverage", CoverageLevel::getDisplayColorsOfCoverageLevel),
+    MODIFIED_FILES(Messages._Baseline_MODIFIED_FILES(), "modifiedFilesCoverage", CoverageLevel::getDisplayColorsOfCoverageLevel),
     /**
      * Difference between the project coverage and the modified file coverage of the current build. Teams can use this delta
      * value to ensure that the coverage of pull requests is better than the whole project coverage.
      */
-    MODIFIED_FILES_DELTA(Messages._Baseline_MODIFIED_FILES_DELTA(), "fileCoverage", CoverageChangeTendency::getDisplayColorsForTendency),
+    MODIFIED_FILES_DELTA(Messages._Baseline_MODIFIED_FILES_DELTA(), "modifiedFilesCoverage", CoverageChangeTendency::getDisplayColorsForTendency),
```
uhafner, line 46: While looking at the results I think I made a mistake here. When I am developing a PR normally I am interested in the coverage of my changes. 

That means we need the totals (this is already working correctly): 
- coverage overall project 
- changed files
- changed lines

But when looking at the delta we need:
- delta overall project - delta overall reference (typically not interesting for large projects)
- delta changed files - delta changed files reference (how is the coverage of the changed files affected)? Currently we check how it changes with respect to the overall coverage.
- delta changed lines: here I am not sure whether it makes more sense to compare vs. the whole project or the changed files. What do you think? Or do we need both?

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,419 @@
...
+                    aggregatedLines = new AggregatedMissingLines(line);
+                }
+                previousLine = line;
+            }
+
+            aggregatedMissingLines.add(aggregatedLines);
+        }
+
+        return aggregatedMissingLines;
+    }
```
uhafner, line 153: Wouldn't it be more object oriented if that code is part of the model? So we should use something like: 

`fileNode.getMissingLines()`, `fileNode.getMissingBranches()`, or `fileNode.getMissingMutations()`?

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,419 @@
...
+                + indirectCoverageChangesLoc;
+    }
+
+    /**
+     * Checks overview regarding the quality gate status.
+     *
+     * @return the markdown string representing the status summary
+     */
+    // TODO: expand with summary of status of each defined quality gate
+    private String getHealthReportSummary() {
```
uhafner, line 236:
```suggestion
    private String getQualityGatesSummary() {
```

uhafner:
I finally managed it to publish a check: https://github.com/uhafner/codingstyle/pull/716/checks?check_run_id=11653892776

uhafner:
Maybe we should also exclude things totally that are not available as in the summary. Otherwise it looks a little bit weird with all the n/a:

Bildschirm­foto 2023-02-28 um 13 14 22

uhafner:
While trying to fix the display of the changes table I am wondering what information we should show actually?

There are several ways: 
- The files table shows all files, the changed files table shows only changed files while displaying the whole source code, the changed lines table shows also all changed files while displaying the changed source code only. 
- The files table shows all files and has a toggle (or filter by column) that shows only modified files. The changed lines table shows all changed files while displaying the changed source code only. 
- Or any other options that make sense?

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageBuildAction.java
fo-code: It added this regarding to formatDelta(...) directly below. What is your suggestion here? Moving both methods to the formatter?

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageChecksPublisher.java
fo-code: Hm the intention of this method is to provide groups of consecutive lines without coverage for the check annotations.
I am not sure if this would be to specific since it is only used here in that special case... In my opinion, it can stay here, but moving it would be okay for me too.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/Baseline.java
fo-code: Yes, that makes sense. 
Regarding the delta changed lines: I guess we should compare only against the modified files coverage. The whole project has modules with higher and with lower coverage. When we compare the changes of a module with lower coverage against the whole project, there will be a negative trend, even if the modified files are covered quiet good compared to the rest of their module. Therefore, comparing against the changed files would keep the context hence it more accurate, in my opinion.

fo-code:
> Maybe we should also exclude things totally that are not available (as done in the summary). Otherwise it looks a little bit weird with all the n/a:
> 
> Bildschirm­foto 2023-02-28 um 13 14 22

Yes, we should do that. n/a only distracts from what is important.

> While trying to fix the display of the changes table I am wondering what information we should show actually?
> 
> There are several ways:
> 
>     * The files table shows all files, the changed files table shows only changed files while displaying the whole source code, the changed lines table shows also all changed files while displaying the changed source code only.
> 
>     * The files table shows all files and has a toggle (or filter by column) that shows only modified files. The changed lines table shows all changed files while displaying the changed source code only.
> 
>     * Or any other options that make sense?
> 
> 
> Possible UI control: Bildschirm­foto 2023-02-28 um 14 08 05

I suggest the second option - that sounds good. By default, we should show only the changed files in my opinion, since most of the time, that's what is interesting - like it is in the screenshot.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageBuildAction.java
uhafner: No, seems that I looked at the wrong class. This is fine here.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageChecksPublisher.java
uhafner: Ok, then lets keep it as it is for now. I am still re-thinking on how to improve the model to make the decision of coverage vs. mutation coverage more object oriented. Currently we have a lot of code that checks if mutation coverage or line or branch coverage (source code painting, etc.).

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageChecksPublisher.java
```diff
@@ -0,0 +1,426 @@
...
+
+        return annotations;
+    }
+
+    private String getAnnotationMessage(final AggregatedMissingLines aggregatedMissingLines) {
+        if (aggregatedMissingLines.startLine == aggregatedMissingLines.endLine) {
+            return Messages.Checks_Annotation_Message_SingleLine(aggregatedMissingLines.startLine);
+        }
+        return Messages.Checks_Annotation_Message_MultiLine(aggregatedMissingLines.startLine,
+                aggregatedMissingLines.endLine);
```
uhafner, line 132: Shouldn't the annotations query for the coverage metric as well? So we have different annotations depending on the metric? And where do you check if the line is covered or not? Should you evaluate the coverage per line as well? So I would expect something like:

- line 23 has no coverage
- line 32 has one missing branch
- line 45 has 4 mutations survived

I'll have a look how I can integrate such kind of method in the model.

I think the best approach would be to have a method in `FileNode` that returns these elements:

```
getMissedLines(final Metric metric)
```

uhafner: I added 3 methods now in https://github.com/uhafner/coverage-model/blob/b3b30707bce09541f2aa976d1a67177ca191c43b/src/main/java/edu/hm/hafner/metric/FileNode.java#L415-L450

uhafner (approved):
I think the PR is now ready to be integrated. I'll track the remaining tasks in #512.

PR #609, opened 2023-03-29
Use buffered reads when recovering coverage result.

y-higuchi:
Use buffered reads when recovering coverage result.

After we started using this plugin, 
we noticed fairly large number of 1 byte reads hitting NFS  server hurting performance.

1 byte reads seems to be coming from [PeekInputStream calls used inside ObjectInputStream](https://github.com/AdoptOpenJDK/openjdk-jdk8u/blob/master/jdk/src/share/classes/java/io/ObjectInputStream.java#L2755) 
directly hitting FileInputStream.

This PR attempts to mitigate the issue by inserting BufferedInputStream to reduce number of small size reads
directly hitting the file system.
 

### DIFF
--- a/plugin/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
+++ b/plugin/src/main/java/io/jenkins/plugins/coverage/CoverageProcessor.java
@@ -1,5 +1,6 @@
 package io.jenkins.plugins.coverage;
 
+import java.io.BufferedInputStream;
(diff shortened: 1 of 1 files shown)


uhafner (approved):
Thanks for your PR.

Please note that this code actually is part of the deprecated code base. This code is not used when you use the new coverage step `recordCoverage`.

Issue #689, opened 2023-05-26
Computing Negative Deltas Doesn't Seem to Work


Jacob-Steele:
### Jenkins and plugins versions report

 Code Coverage API Version 4.4.0 

### What Operating System are you using (both controller, and any agents involved in the problem)?

Linux Ubuntu Agent and Controller

### Reproduction steps
1. Have a project setup where code coverage is dropping when a pull request is opened.
2. Run `recordCoverage` with the following configuration
```
recordCoverage(
        checksAnnotationScope: 'MODIFIED_LINES', checksName: 'Code Coverage',
        failOnError: true, name: 'Code Coverage',
        qualityGates: [
            [baseline: 'PROJECT_DELTA', metric: 'LINE', threshold: -1.0, criticality: 'UNSTABLE'],
            [baseline: 'PROJECT_DELTA', metric: 'BRANCH', threshold: -1.0, criticality: 'UNSTABLE']
        ],
        sourceDirectories: [[path: 'src']],
        scm: 'code-repo', sourceCodeRetention: 'LAST_BUILD',
        tools: [[parser: 'JACOCO', pattern: '**/reports/jacoco/**/jacocoTestReport.xml']]
)
```

### Expected Results

```
13:21:49  [Coverage] Evaluating quality gates
13:21:49  [Coverage] -> Some quality gates have been missed: overall result is UNSTABLE
13:21:49  [Coverage] -> Details for each quality gate:
13:21:49  [Coverage] -> [Overall project (difference to reference job) - Line Coverage]: ≪Success≫ - (Actual value: +0.33%, Quality gate: -1.00)
13:21:49  [Coverage] -> [Overall project (difference to reference job) - Branch Coverage]: ≪Unstable≫ - (Actual value: -1.37%, Quality gate: -1.00)
```

### Actual Results

```
13:21:49  [Coverage] Evaluating quality gates
13:21:49  [Coverage] -> All quality gates have been passed
13:21:49  [Coverage] -> Details for each quality gate:
13:21:49  [Coverage] -> [Overall project (difference to reference job) - Line Coverage]: ≪Success≫ - (Actual value: +0.33%, Quality gate: -1.00)
13:21:49  [Coverage] -> [Overall project (difference to reference job) - Branch Coverage]: ≪Success≫ - (Actual value: -1.37%, Quality gate: -1.00)
```

### Anything else?

[...]

PR #703, opened 2023-06-04
Convert delta fractions to percentages when used in quality gates

uhafner:
Delta values are internally stored as fractions, but a user expects them as percentages when they are displayed or used in thresholds. So values in the range [-1, 1] should be converted to percentages in the range [-100%, 100%] before passed to the quality gate evaluation.

Fixes https://github.com/jenkinsci/code-coverage-api-plugin/issues/689


Jacob-Steele:
Thank you so much @uhafner for the quick fix on this!

Issue #731, opened 2023-07-16
Remote API to connect additional code review tools


fo-code:
### What feature do you want to see added?

I would like to show the calculated coverage results in  additional code review tools except GitHub so the user can review the code and instantly see if the modified lines are covered even without navigating to Jenkins.
This would be a big enhancement since it further decreases the effort to see required coverage information hence makes it more attractive for developers. Also, additional tools can be combined with this plugin.

More precisely, we want to implement a REST API that provides the coverage information of modified lines for each modified file.

My colleague and I would also implement this, but we are not sure how to do it in the best way. 
There is the CoverageViewModel that has a getApi() method that already offers more general information.
How can we extend the plugin with an additional API as described above and which steps would be required in general?

@uhafner what do you think about this idea and could you please help us with our questions?

uhafner:
There is some [documentation](https://www.jenkins.io/doc/developer/handling-requests/json/) in our developer guide.

In the warnings plugin I am using several API access points:
https://github.com/jenkinsci/warnings-ng-plugin/tree/master/plugin/src/main/java/io/jenkins/plugins/analysis/core/restapi

You can even write a help file that shows the available end points:

https://github.com/jenkinsci/warnings-ng-plugin/blob/master/plugin/src/main/resources/io/jenkins/plugins/analysis/core/restapi/AnalysisResultApi/_api.jelly

So you can start the API end point in either a new URL or a sub page of `coverage`.

fo-code:
Thanks a lot for helping! We will create a draft PR with the initial code and than enhance it step by step here.

PR #737, opened 2023-07-31
Remote API to connect additional code review tools

fo-code:
Adds a REST API that provides the coverage information of modified lines for each modified file - see #731 .


plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/ChangedLinesModel.java
```diff
@@ -0,0 +1,57 @@
...
+
+import java.util.Objects;
+
+public class ChangedLinesModel {
+    private int startLine;
+    private int endLine;
+    private Type type;
+
+    public ChangedLinesModel(final int start_line, final int end_line, final Type type) {
+        this.setStartLine(start_line);
```
fo-code, line 11: We should not call other methods from a constructor that can be overwritten. This could unexpectedly change the constructor's behavior in subclass. Directly set the attributes and make them final. 
Hence, we do not need setters here,

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/ChangedLinesModel.java
```diff
@@ -0,0 +1,57 @@
...
+        this.setEndLine(end_line);
+        this.setType(type);
+    }
+
+    @Override
+    public boolean equals(final Object o) {
```
fo-code, line 17: These should be added in the end of the file since the getters are more of interest than equals/hashcode

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/Type.java
```diff
@@ -0,0 +1,7 @@
+package io.jenkins.plugins.coverage.metrics.model;
+
+public enum Type {
```
fo-code, line 3: Type is to generic as class name in my opinion. The reference to coverage is missing.
Maybe LineCoverageType?
A look at the used [Coverage-Model](https://github.com/jenkinsci/coverage-model) might be also helpful to see what is already implemented and how the model classes are named.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/Side.java
```diff
@@ -0,0 +1,6 @@
+package io.jenkins.plugins.coverage.metrics.model;
+
+public enum Side {
```
fo-code, line 3: I am not sure if this is really necessary for coverage since always the right side / new code is of interest.
I think this can be removed.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/LineCoverageViewModel.java
```diff
@@ -0,0 +1,76 @@
...
+@ExportedBean
+public class LineCoverageViewModel {
+
+    private final List<FileWithChangedLinesCoverageModel> filesWithChangedLines;
+
+    public LineCoverageViewModel(final Node node) {
+        this.filesWithChangedLines = getFilesWithChangedLines(node);
+    }
+
+    public static List<FileWithChangedLinesCoverageModel> getFilesWithChangedLines(final Node node) {
```
fo-code, line 25: The business logic should not be part of the data class.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/LineCoverageViewModel.java
```diff
@@ -0,0 +1,76 @@
...
+import io.jenkins.plugins.coverage.metrics.model.FileWithChangedLinesCoverageModel;
+import io.jenkins.plugins.coverage.metrics.model.Type;
+
+@ExportedBean
+public class LineCoverageViewModel {
+
+    private final List<FileWithChangedLinesCoverageModel> filesWithChangedLines;
+
+    public LineCoverageViewModel(final Node node) {
+        this.filesWithChangedLines = getFilesWithChangedLines(node);
```
fo-code, line 22: The constructor can get the list of FileWithChangedLinesCoverageModel and directly initialize the attribute here so it can be final.

plugin/src/test/java/io/jenkins/plugins/coverage/metrics/steps/LineCoverageViewTest.java
```diff
@@ -0,0 +1,75 @@
...
+import io.jenkins.plugins.coverage.metrics.model.FileWithChangedLinesCoverageModel;
+import io.jenkins.plugins.coverage.metrics.model.Type;
+
+import static org.hamcrest.MatcherAssert.*;
+import static org.hamcrest.core.Is.*;
+
+/**
+ * TODO: JavaDoc me
+ */
+class LineCoverageViewTest extends AbstractModifiedFilesCoverageTest {
```
fo-code, line 20: The name of this test class does not really describe what is tested. After the business logic is seperated from the data classes, a test for specificely this business class can be written.
Example: MyBusinessClass.java -> MyBusinessClassTest.java

plugin/src/test/java/io/jenkins/plugins/coverage/metrics/steps/LineCoverageViewTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+
+        var fileOneLinesList = createListOfChangedLines(Type.COVERED, 15, 16, 21, 22);
+        fileOneLinesList.addAll(createListOfChangedLines(Type.MISSED, 35, 36));
+        fileOneLinesList.addAll(createListOfChangedLines(Type.PARTRIALLY_COVERED, 20, 20));
+
+        var fileOne = new FileWithChangedLinesCoverageModel("test/example/Test1.java");
+        fileOne.setListOfChangedLines(fileOneLinesList);
+
+        assertThat(filesWithChangedLines, IsCollectionContaining.hasItem(fileOne));
```
fo-code, line 60: AssertJ would be preferable here instead of Hamcrest.
The assertion here is also duplicated, the assertions above are enough.
Also, something about the equals methods must not be correct since the assertion matches even if there are only three elements in fileOne and the first one combines two of the tested filesWithChangedLines.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/FileWithChangedLinesCoverageModel.java
```diff
@@ -0,0 +1,55 @@
...
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof FileWithChangedLinesCoverageModel)) {
+            return false;
+        }
+        FileWithChangedLinesCoverageModel file = (FileWithChangedLinesCoverageModel) o;
+        return Objects.equals(getRelativePath(), file.getRelativePath())
+                //&& Objects.equals(getListOfChangedLines(), file.getListOfChangedLines());
```
fo-code, line 24: Why?

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/FileWithChangedLinesCoverageModel.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.List;
+import java.util.Objects;
+
+import org.kohsuke.stapler.export.Exported;
+import org.kohsuke.stapler.export.ExportedBean;
+
+@ExportedBean
+public class FileWithChangedLinesCoverageModel {
+    private String relativePath;
```
fo-code, line 11: These should be private final fields set be the constructor.

fo-code (commented):
Good initial implementation - I added a few comments.
Please also have a look the the GitHub Checks and fix the Checkstyle, PMD, ... Warnings.

Generally, the model classes should not contain "Model" in their name in my opinion. Also, I think we can find better names for the classes to better match their use case - we should rethink that. "Changed Lines" here for example is always described as "Modified Lines". We should keep this consistent.

plugin/src/test/java/io/jenkins/plugins/coverage/metrics/steps/CoverageApiUtilTest.java
```diff
@@ -13,10 +13,10 @@
 import static org.assertj.core.api.Assertions.*;
 
 /**
- * Test to assert that all modified lines and their respective coverage types are correctly extracted from the
- * coverage tree created by the createCoverageTree() method, using the getFilesWithModifiedLines() method.
+ * Test to assert that all modified lines and their respective coverage types are correctly extracted from the coverage
+ * tree created by the createCoverageTree() method, using the calculateFilesWithModifiedLines() method in {@link CoverageApiUtil}.
```
fo-code, line 17: Here the method can be linked directly

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageApiUtil.java
```diff
@@ -10,57 +10,64 @@
...
-public class LineCoverageViewModel {
+public class CoverageApiUtil {
 
-    private LineCoverageViewModel() {
+    private CoverageApiUtil() {
         //this is a utility class
     }
 
     /**
      * Static method to extract modified lines of code and their respective coverage status from passed Node object.
```
fo-code, line 23: "Static method" is unnecessary information

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/LineCoverageType.java
```diff
@@ -0,0 +1,10 @@
+package io.jenkins.plugins.coverage.metrics.model;
+
+/**
+ * Enum to define the type of coverage assigned to any ModifiedLinesBlock object.
+ */
+public enum LineCoverageType {
+    COVERED,
+    MISSED,
+    PARTRIALLY_COVERED
```
fo-code, line 9: Typo

fo-code (commented):
Looks great - only a few comments.

Another idea: 
we can extend the FileWithModifiedLines with indirect coverage changes aswell (see FileNode#getIndirectCoverageChanges();).
This shows which lines have coverage changes caused by changed test cases. Would be also helpful i guess.
This requires another LineCoverageType: INDIRECT_CHANGE. Should be easy to include since the calculation is the same as for the other types.

fo-code:
@uhafner Nonetheless there are still a few tasks to do, the API now is in a final state.
What is your opinion on this? I would like to wait with merging until we have your confirmation. :)

plugin/src/test/java/io/jenkins/plugins/coverage/metrics/steps/CoverageApiUtilTest.java
aelfric-m: Done.

plugin/src/test/java/io/jenkins/plugins/coverage/metrics/AbstractModifiedFilesCoverageTest.java
```diff
@@ -90,20 +92,23 @@ protected Node createCoverageTree() {
...
      * Gets the path of the test file with modified lines.
+     * @return TEST_FILE_MODIFIED_PATH returns the relative path of a modified test file.
      */
     protected String getPathOfFileWithModifiedLines() {
         return TEST_FILE_MODIFIED_PATH;
     }
 
     /**
      * Gets the name of the test file without modified lines.
+     * @return TEST_FILE_NOT_MODIFIED returns the name of a file with no modified code lines.
```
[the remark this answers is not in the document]
aelfric-m: Done.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageApiUtil.java
```diff
@@ -10,57 +10,64 @@
...
+     *
+     * @param modifiedLines
+     *         list containing the integer numbers of modified lines.
+     * @param modifiedLinesBlocks
+     *         list containing modifiedLinesBlock objects.
+     * @param type
+     *         type of coverage pertaining to each line of code (COVERED, MISSED, or PARTIALLY_COVERED)
      */
-    public static void getModifiedLineBlocks(final List<Integer> modifiedLines,
+    public static void calculateModifiedLineBlocks(final List<Integer> modifiedLines,
```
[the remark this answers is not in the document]
aelfric-m: Done. Also corrected a type-o in PARTIALLY_COVERED.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageApiUtil.java
aelfric-m: Removed.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageApiUtil.java
```diff
@@ -10,57 +10,64 @@
 import io.jenkins.plugins.coverage.metrics.model.LineCoverageType;
 import io.jenkins.plugins.coverage.metrics.model.ModifiedLinesBlock;
 
-//TODO: rename to util
 /**
  * Server side model that provides data for the details of line coverage results in modified lines.
```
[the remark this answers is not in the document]
aelfric-m: corrected.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageApiUtil.java
```diff
@@ -10,57 +10,64 @@
...
      */
     public static List<FileWithModifiedLines> getFilesWithModifiedLines(final Node node) {
-        var filesWithModifiedLinesList = new ArrayList<FileWithModifiedLines>();
+        var filesWithModifiedLines = new ArrayList<FileWithModifiedLines>();
 
-        for (FileNode fileNode :  node.filterByModifiedLines().getAllFileNodes()) {
-            var listOfMissedLines = new ArrayList<>(fileNode.getMissedLines());
-            var listOfPartialLines = new ArrayList<>(fileNode.getPartiallyCoveredLines().keySet());
-            List<Integer> listOfCoveredLines = new ArrayList<>();
+        for (FileNode fileNode : node.filterByModifiedLines().getAllFileNodes()) {
```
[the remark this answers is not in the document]
aelfric-m: Done.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageApiUtil.java
```diff
@@ -10,57 +10,64 @@
...
+            var missedLines = new ArrayList<>(fileNode.getMissedLines());
+            var partiallyCoveredLines = new ArrayList<>(fileNode.getPartiallyCoveredLines().keySet());
+            List<Integer> coveredLines = new ArrayList<>();
 
             int i = 0;
-            for (Integer a:fileNode.getLinesWithCoverage()) {
+            for (Integer a : fileNode.getLinesWithCoverage()) {
                 if (fileNode.getCoveredCounters()[i] > 0 && fileNode.getMissedCounters()[i] == 0) {
```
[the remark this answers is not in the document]
aelfric-m: Corrected. Changed variable name to "line" and removed the counter variable i.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/ModifiedLinesBlock.java
```diff
@@ -6,8 +6,9 @@
 import org.kohsuke.stapler.export.ExportedBean;
 
 /**
- * Model class containing data pertained to consecutive lines of modified code. Each object possesses a starting and ending
- * line number and the type of coverage of the block. Each object is associated with a FileWithModifiedLines object.
+ * Model class containing data pertained to consecutive lines of modified code. Each object possesses a starting and
+ * ending line number and the type of coverage of the block. Each object is associated with a FileWithModifiedLines
+ * object.
```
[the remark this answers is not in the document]
aelfric-m: Done.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/FileWithModifiedLines.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.List;
+import java.util.Objects;
+
+import org.kohsuke.stapler.export.Exported;
+import org.kohsuke.stapler.export.ExportedBean;
+
+/**
+ * Model class containing data pertaining to files with modified lines of code.
+ * Each object contains a relative file path and a list of ModifiedLinesBlock objects.
```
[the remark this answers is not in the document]
aelfric-m: Done.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/LineCoverageType.java
aelfric-m: Done

plugin/src/test/java/io/jenkins/plugins/coverage/metrics/steps/CoverageApiUtilTest.java
```diff
@@ -0,0 +1,62 @@
...
+        var fileOneLinesList = createListOfModifiedLines(LineCoverageType.COVERED, 15, 16, 21, 22);
+        fileOneLinesList.addAll(createListOfModifiedLines(LineCoverageType.MISSED, 35, 36));
+        fileOneLinesList.addAll(createListOfModifiedLines(LineCoverageType.PARTRIALLY_COVERED, 20, 20));
+
+        var fileOne = new FileWithModifiedLines("test/example/Test1.java", fileOneLinesList);
+        var fileTwo = new FileWithModifiedLines("fileTwo", null);
+        var checkTrue = fileOne.equals(fileOne);
+        var checkFalse = fileOne.equals(fileTwo);
+
+        assertThat(checkTrue).isTrue();
```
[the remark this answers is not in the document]
aelfric-m: Done.

plugin/src/test/java/io/jenkins/plugins/coverage/metrics/steps/CoverageApiUtilTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Test
+    void testOverriddenEqualsMethod() {
+        var fileOneLinesList = createListOfModifiedLines(LineCoverageType.COVERED, 15, 16, 21, 22);
+        fileOneLinesList.addAll(createListOfModifiedLines(LineCoverageType.MISSED, 35, 36));
+        fileOneLinesList.addAll(createListOfModifiedLines(LineCoverageType.PARTRIALLY_COVERED, 20, 20));
+
+        var fileOne = new FileWithModifiedLines("test/example/Test1.java", fileOneLinesList);
+        var fileTwo = new FileWithModifiedLines("fileTwo", null);
+        var checkTrue = fileOne.equals(fileOne);
```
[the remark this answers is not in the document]
aelfric-m: If I remove this, the line of code in the equals() method that returns true if the same object is passed to it as parameter, will not be covered by tests. I agree, that this doesn't look good, so am open for suggestions.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/FileWithModifiedLines.java
```diff
@@ -0,0 +1,56 @@
...
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof FileWithModifiedLines)) {
+            return false;
+        }
+        FileWithModifiedLines file = (FileWithModifiedLines) o;
+        return Objects.equals(this.getRelativePath(), file.getRelativePath())
+                && this.getListOfModifiedLines().containsAll(file.getListOfModifiedLines())
```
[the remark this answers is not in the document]
aelfric-m: Done.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/ModifiedLinesBlock.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof ModifiedLinesBlock)) {
+            return false;
+        }
```
uhafner, line 42: Can you use the same convention for equals as all other classes? Typically `getClass` is used in equals.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/FileWithModifiedLines.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Objects;
+
+import org.kohsuke.stapler.export.Exported;
+import org.kohsuke.stapler.export.ExportedBean;
+
+/**
+ * Model class containing data pertaining to files with modified lines of code. Each object contains a relative file
+ * path and a list of {@link ModifiedLinesBlock} objects.
+ */
+@ExportedBean
```
uhafner, line 13: Are you planning to use these classes outside of the API?

Otherwise I would recommend to create a new package (restapi, remoting, api, ...) and move all classes there. Then we can move the *Api classes to the new package as well. 

This would allow to make all internal classes package private as well.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/FileWithModifiedLines.java
```diff
@@ -0,0 +1,58 @@
...
+    private final List<ModifiedLinesBlock> listOfModifiedLines;
+
+    /**
+     * Constructor for the FileWithModifiedLines class.
+     *
+     * @param relativePath
+     *         the relative path of the file
+     * @param listOfModifiedLines
+     *         all modified lines inside the file
+     */
```
uhafner, line 25: Either make package private (if possible, see above) and remove the comment. Or use something like:
```suggestion
    /**
     * Creates a new instance of {@link FileWithModifiedLines}.
     *
     * @param relativePath
     *         the relative path of the file
     * @param listOfModifiedLines
     *         all modified lines inside the file
     */
```

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/LineCoverageType.java
```diff
@@ -0,0 +1,10 @@
+package io.jenkins.plugins.coverage.metrics.model;
+
+/**
+ * Enum to define the type of coverage assigned to any ModifiedLinesBlock object.
```
uhafner, line 4:
```suggestion
 * Defines the type of line coverage assigned to any {@link ModifiedLinesBlock} object.
```

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/ModifiedLinesBlock.java
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Constructor for this class.
+     *
+     * @param startLine
+     *         number of the first line of code associated with this block.
+     * @param endLine
+     *         number of the last line of code associated with this block.
+     * @param type
+     *         type of coverage.
+     */
```
uhafner, line 28: See above. You should create a template in your IDE for such comments

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageBuildAction.java
```diff
@@ -609,6 +609,7 @@ protected String getBuildResultBaseName() {
 
     @Override
     public CoverageViewModel getTarget() {
+
```
uhafner, line 612: Please remove unrelated formatting changes

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageApiUtil.java
```diff
@@ -0,0 +1,90 @@
...
+import edu.hm.hafner.coverage.Node;
+
+import io.jenkins.plugins.coverage.metrics.model.FileWithModifiedLines;
+import io.jenkins.plugins.coverage.metrics.model.LineCoverageType;
+import io.jenkins.plugins.coverage.metrics.model.ModifiedLinesBlock;
+
+/**
+ * Server side model that provides data for the details of line coverage results of modified lines.
+ */
+public class CoverageApiUtil {
```
uhafner, line 16: This class and methods should be actually part of the API class. 

Then the constructor of `LineCoverageApi` works on a `Node`. You can make the method package private to test it there.

uhafner (changes requested):
Thanks for your pull request! 

I will finish the review later this week (tests).

plugin/pom.xml
```diff
@@ -202,6 +202,12 @@
     </dependency>
 
     <!-- Test Dependencies -->
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <version>3.23.1</version>
+      <scope>test</scope>
+    </dependency>
```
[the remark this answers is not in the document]
aelfric-m: Removed.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageApi.java
```diff
@@ -136,7 +136,10 @@ public Result getOverallResult() {
 
         @Exported(inline = true)
         public Collection<QualityGateItemApi> getResultItems() {
-            return qualityGateResult.getResultItems().stream().map(QualityGateItemApi::new).collect(Collectors.toList());
+            return qualityGateResult.getResultItems()
+                    .stream()
+                    .map(QualityGateItemApi::new)
+                    .collect(Collectors.toList());
```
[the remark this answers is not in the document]
aelfric-m: Removed.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageViewModel.java
```diff
@@ -73,7 +73,8 @@ public class CoverageViewModel extends DefaultAsyncTableContentProvider implemen
     private static final String INFO_MESSAGES_VIEW_URL = "info";
 
     private static final ElementFormatter FORMATTER = new ElementFormatter();
-    private static final Set<Metric> TREE_METRICS = Set.of(Metric.LINE, Metric.INSTRUCTION, Metric.BRANCH, Metric.MUTATION);
+    private static final Set<Metric> TREE_METRICS = Set.of(Metric.LINE, Metric.INSTRUCTION, Metric.BRANCH,
+            Metric.MUTATION);
```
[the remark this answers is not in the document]
aelfric-m: Removed.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageViewModel.java
```diff
@@ -347,7 +357,7 @@ public String getSourceCode(final String fileHash, final String tableId) {
         if (targetResult.isPresent()) {
             try {
                 Node fileNode = targetResult.get();
-                return readSourceCode((FileNode)fileNode, tableId);
+                return readSourceCode((FileNode) fileNode, tableId);
```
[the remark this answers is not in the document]
aelfric-m: Removed.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageBuildAction.java
aelfric-m: Removed.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/ModifiedLinesBlock.java
aelfric-m: Changed.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/model/FileWithModifiedLines.java
aelfric-m: Done. Files have been made package private if possible, otherwise the correct format is being used.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/steps/CoverageApiUtil.java
aelfric-m: We have been on the fence about this. The logic is now part of the API class again.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/restapi/CoverageApi.java
```diff
@@ -1,4 +1,4 @@
-package io.jenkins.plugins.coverage.metrics.steps;
+package io.jenkins.plugins.coverage.metrics.restapi;
```
uhafner, line 1: You need to add that new package to https://github.com/jenkinsci/code-coverage-api-plugin/blob/master/plugin/src/test/resources/design.puml to get the architecture tests running

plugin/src/main/resources/io/jenkins/plugins/coverage/metrics/restapi/_api.jelly
```diff
@@ -0,0 +1,17 @@
+<?jelly escape-by-default='true'?>
```
uhafner, line 1: I'm not sure if this is the right package for that file? I can't see the content. Do you see that text?

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/restapi/CoverageApi.java
aelfric-m: Done.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/restapi/ModifiedLineCoverageApi.java
```diff
@@ -0,0 +1,101 @@
...
+import edu.hm.hafner.coverage.Node;
+
+import org.kohsuke.stapler.export.Exported;
+import org.kohsuke.stapler.export.ExportedBean;
+
+/**
+ * Remote API to list the details of modified line coverage results.
+ */
+@ExportedBean
+public class ModifiedLineCoverageApi {
```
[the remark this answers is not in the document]
aelfric-m: This is true. The name still seems a bit clunky, but will do for now.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/gerrit/GerritCoverageController.java
```diff
@@ -0,0 +1,111 @@
...
+import hudson.model.ParametersAction;
+import hudson.model.Run;
+import hudson.model.UnprotectedRootAction;
+
+import io.jenkins.plugins.coverage.metrics.restapi.ModifiedLinesCoverageApi;
+import io.jenkins.plugins.coverage.metrics.steps.CoverageBuildAction;
+import io.jenkins.plugins.util.JenkinsFacade;
+
+@Extension
+public class GerritCoverageController implements UnprotectedRootAction {
```
uhafner, line 26: What is this class for? It does not belong to this plugin and should be moved somewhere else.

fo-code: This is required for using the coverage API in Gerrit since there we have no information about the Jenkins build number that is required for accessing the API. 
We thought some other people could use it aswell, maybe on a different location where others could also add similar extension for other tools for which the API does not work out of the box? 

If this does not belong to this plugin in your opinion, we can only use it in our fork of course. :)

uhafner: Yes, I would prefer to get that file removed. It would make sense that you put it in a small additional plugin that contains the Gerrit and Jenkins bridge.

plugin/src/main/resources/io/jenkins/plugins/coverage/metrics/restapi/_api.jelly
fo-code: Done. Has been in the wrong directory indeed.

fo-code:
@uhafner I extended the GitForensicsITest with a call to the new API to verify the resulting JSON. I added that in that test class since it offers a special setup where modified code lines as well as related coverage data is provided. Both is required for a decent test scenario here.
Also, I finished the implementation and added a description of the new API to the README.

From my side we can merge this PR, if you don't have any further suggestions.

plugin/src/main/resources/io/jenkins/plugins/coverage/metrics/restapi/ModifiedLinesCoverageApi/_api.jelly
```diff
@@ -0,0 +1,17 @@
+<?jelly escape-by-default='true'?>
```
uhafner, line 1: This file needs to be moved to `[...]/coverage/metrics/restapi/CoverageApi/_api.jelly`. It should be shown when the user clicks on the API link on the `coverage` view.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/restapi/ModifiedLinesCoverageApi.java
```diff
@@ -0,0 +1,97 @@
...
+        for (FileNode fileNode : node.filterByModifiedLines().getAllFileNodes()) {
+            var missedLines = new ArrayList<>(fileNode.getMissedLines());
+            var partiallyCoveredLines = new ArrayList<>(fileNode.getPartiallyCoveredLines().keySet());
+            var coveredLines = new ArrayList<Integer>();
+
+            for (Integer line : fileNode.getLinesWithCoverage()) {
+                if (fileNode.getMissedOfLine(line) == 0 && fileNode.getCoveredOfLine(line) > 0) {
+                    coveredLines.add(line);
+                }
+            }
```
uhafner, line 45: Wouldn't it make sense to move the method to the coverage model? (This can be done after the merge...)

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/restapi/ModifiedLinesCoverageApi.java
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Finds all files with modified lines of code in a passed {@link Node} object.
+     *
+     * @param node
+     *         containing the file tree.
+     *
+     * @return a list of {@link FileWithModifiedLines} objects.
+     */
+    private static List<FileWithModifiedLines> createListOfFilesWithModifiedLines(final Node node) {
```
uhafner, line 33:
```suggestion
    private List<FileWithModifiedLines> createFilesWithModifiedLines(final Node node) {
```

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/restapi/ModifiedLinesCoverageApi.java
```diff
@@ -0,0 +1,97 @@
...
+                if (fileNode.getMissedOfLine(line) == 0 && fileNode.getCoveredOfLine(line) > 0) {
+                    coveredLines.add(line);
+                }
+            }
+
+            var modifiedLinesBlocks = new TreeSet<ModifiedLinesBlock>();
+
+            calculateModifiedLineBlocks(coveredLines, modifiedLinesBlocks, LineCoverageType.COVERED);
+            calculateModifiedLineBlocks(missedLines, modifiedLinesBlocks, LineCoverageType.MISSED);
+            calculateModifiedLineBlocks(partiallyCoveredLines, modifiedLinesBlocks, LineCoverageType.PARTIALLY_COVERED);
```
uhafner, line 51: It is somehow untypical in Java to use a parameter as output value. Why don't you use a return value and add the result to the list?

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/restapi/ModifiedLinesCoverageApi.java
```diff
@@ -0,0 +1,97 @@
...
+     *
+     * @param modifiedLines
+     *         list containing the integer numbers of modified lines.
+     * @param modifiedLinesBlocks
+     *         list containing modifiedLinesBlock objects.
+     * @param type
+     *         type of coverage pertaining to each line of code ({@link LineCoverageType#COVERED},
+     *         {@link LineCoverageType#MISSED}, or {@link LineCoverageType#PARTIALLY_COVERED})
+     */
+    private static void calculateModifiedLineBlocks(final List<Integer> modifiedLines,
```
uhafner, line 72:
```suggestion
    private void calculateModifiedLineBlocks(final List<Integer> modifiedLines,
```

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/restapi/ModifiedLinesCoverageApi.java
```diff
@@ -0,0 +1,97 @@
...
+
+                var modifiedLinesBlock = new ModifiedLinesBlock(currentLine, modifiedLines.get(i), type);
+                modifiedLinesBlocks.add(modifiedLinesBlock);
+
+                if (i < modifiedLines.size() - 1) {
+                    currentLine = modifiedLines.get(i + 1);
+                }
+            }
+        }
+    }
```
uhafner, line 91: I'm not sure if this is relevant: the result is not accurate as typically not all source code lines have a coverage counter (comments, empty lines). As a user I would assume that those not measured lines would not create a new lines block. See https://github.com/jenkinsci/coverage-model/blob/67df3455cb8577a872eecef5c8daae0564c445a7/src/main/java/edu/hm/hafner/coverage/FileNode.java#L576 for an alternative implementation of this problem.

uhafner (changes requested):
Almost ready for merge, thanks for updating!

plugin/src/test/java/io/jenkins/plugins/coverage/metrics/restapi/ModifiedLinesCoverageApiTest.java
```diff
@@ -0,0 +1,61 @@
...
+import io.jenkins.plugins.coverage.metrics.AbstractModifiedFilesCoverageTest;
+
+import static org.assertj.core.api.Assertions.*;
+
+/**
+ * Tests {@link ModifiedLinesCoverageApi}.
+ */
+
+class ModifiedLinesCoverageApiTest extends AbstractModifiedFilesCoverageTest {
+    //TODO: write shouldProvideRemoteApi() test
```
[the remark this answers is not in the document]
fo-code: Yes, just forgot this

plugin/src/main/resources/io/jenkins/plugins/coverage/metrics/restapi/ModifiedLinesCoverageApi/_api.jelly
fo-code: Ah, I see. Thanks for the hint.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/restapi/ModifiedLinesCoverageApi.java
fo-code: Yes true, I implemented another solution that should fix this problem.

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/restapi/ModifiedLinesCoverageApi.java
fo-code: Yes I think so :)

plugin/src/main/java/io/jenkins/plugins/coverage/metrics/restapi/ModifiedLinesCoverageApi.java
fo-code: I improved the implementation like this:
- the API will provide only coverage information for modified lines (as described, has been wrong before)
- lines without coverage will be included in a block if:
  - the lines with coverage information before and after have the same coverage type
  - the lines without coverage information have also been modified (else, there is too much not required information since the user is only interested in modified lines here, in my opinion)
- I added Unit tests to cover this behavior

fo-code:
@uhafner I reworked the implementation and included your comments. Now everything works like discussed and is also tested.

The only problem is there are test failures regarding docker images and I think they are not caused by this pull request, so I do not know how to fix it.

uhafner (approved):
Thanks!

uhafner:
I am rerunning the checks and will merge if everything is back to green...

Issue #785, opened 2023-10-09
Exception for Cobertura reports that contain duplicate methods


dannygueta:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
 Jenkins: 2.414.2
OS: Linux - 5.10.192-183.736.amzn2.x86_64
Java: 11.0.20.1 - Eclipse Adoptium (OpenJDK 64-Bit Server VM)
---
ace-editor:1.1
analysis-model-api:11.9.0
ansicolor:1.0.4
antisamy-markup-formatter:162.v0e6ec0fcfcf6
apache-httpcomponents-client-4-api:4.5.14-208.v438351942757
authentication-tokens:1.53.v1c90fd9191a_b_
badge:1.9.1
blueocean:1.27.8
blueocean-bitbucket-pipeline:1.27.8
blueocean-commons:1.27.8
blueocean-config:1.27.8
blueocean-core-js:1.27.8
blueocean-dashboard:1.27.8
blueocean-display-url:2.4.2
blueocean-events:1.27.8
blueocean-git-pipeline:1.27.8
blueocean-github-pipeline:1.27.8
blueocean-i18n:1.27.8
blueocean-jwt:1.27.8
blueocean-personalization:1.27.8
blueocean-pipeline-api-impl:1.27.8
blueocean-pipeline-editor:1.27.8
blueocean-pipeline-scm-api:1.27.8
blueocean-rest:1.27.8
blueocean-rest-impl:1.27.8
blueocean-web:1.27.8
bootstrap4-api:4.6.0-6
bootstrap5-api:5.3.2-1
bouncycastle-api:2.29
branch-api:2.1128.v717130d4f816
build-with-parameters:76.v9382db_f78962
caffeine-api:3.1.8-133.v17b_1ff2e0599
checks-api:2.0.2
cloudbees-bitbucket-branch-source:848.v42c6a_317eda_e
cloudbees-folder:6.848.ve3b_fd7839a_81
cobertura:1.17
code-coverage-api:4.8.0
command-launcher:107.v773860566e2e
commons-lang3-api:3.13.0-62.v7d18e55f51e2
commons-text-api:1.10.0-78.v3e7b_ea_d5a_fe1
config-file-provider:959.vcff671a_4518b_
configuration-as-code:1714.v09593e830cfa
credentials:1293.vff276f713473
credentials-binding:636.v55f1275c7b_27
[...]

PR #788, opened 2023-10-11
Add a new option to ignore parser errors

uhafner:
For more details, please see https://github.com/jenkinsci/coverage-model/pull/39.

Fixes https://github.com/jenkinsci/code-coverage-api-plugin/issues/785

### DIFF
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -33,7 +33,7 @@
     <testcontainers.version>1.19.1</testcontainers.version>
     <job-dsl.version>1.84</job-dsl.version>
 
-    <coverage-model.version>0.25.0</coverage-model.version>
+    <coverage-model.version>0.27.0</coverage-model.version>
     <git-forensics.version>2.0.0</git-forensics.version>
     <prism-api.version>1.29.0-8</prism-api.version>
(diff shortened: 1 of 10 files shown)


medianick:
We ran into this issue and I adopted this new `ignoreParsingErrors: true` option in our `recordCoverage` calls. That mostly solved the issue we were encountering, but we now transiently get errors like this (despite the setting):

```
There is already a child [CLASS] My.Redacted.Class`2 <0> with the name My.Redacted.Class`2 in [FILE] RedactedFile.cs <1, LINE: 88.89% (32/36)>
```

I have to imagine it is the result of the disambiguation issue documented at https://github.com/danielpalme/ReportGenerator/issues/630 but we don't seem to have any way around it, short of rerunning the build and hoping it works. (We could also add try/catch logic around `recordCoverage` but depend on its output as a status check.)

I'm working on obtaining the specific coverage reports that seem to cause this, but as I noted, it's transient -- rerunning a build of the same commit will sometimes fail and sometimes work.

uhafner:
Currently I catch errors on duplicate methods only. I should extend this to duplicate classes as well.

